PR #2, opened 2015-02-09
'request' is a production dependency

dwrensha:
If I install dependencies like this:

```
$ NODE_ENV=production npm install
```

this component crashes with an error about being unable to find 'request'.


jpallen:
Thanks, good catch!

jpallen:
Btw, if you're looking at upgrading ShareLaTeX in Sandstorm, I'm about a day or two away from releasing a new 'point version' of ShareLaTeX which you may want to wait for.

dwrensha:
Yep, I've updated our port of ShareLaTeX. I just rebased today to include the 0.1.2 release. You can try it here: https://demo.sandstorm.io/appdemo/5vuv7v0w7gu20z72m78n83rx9qqtqpmtk32f39823wh967z226qh

In fact, this demo link will continue to work, as Asheesh explains in this blog post:
https://blog.sandstorm.io/news/2015-02-06-app-demo.html

jpallen:
Cool stuff :) Enjoying watching your progress!

PR #3, opened 2015-05-11
Automatically compile files and restart service when files change

dpasqualin:
When I saw [this](https://github.com/sharelatex/sharelatex/wiki/Setting-up-a-Development-Environment#recompiling-coffeescript---js) I though we could do better. grunt-contrib-watch watch files for change and run tasks when they change. In this case, recompile coffee files and restart the service.

I had to replace grunt-execute with grunt-express-server, because the latter can spawn the node process in background, allowing the watch to regain control of the grunt process and do its business.

If you like the idea I could replicate it in all sharelatex repositories, finally adding this option in the main Gruntfile on sharelatex repository, maybe facillitating the development.


dpasqualin:
Oops, I have to fix one thing or two here. I'm gonna close it for a while. But feel free to comment if you think its a nice thing to have.

jpallen:
We need to standardize the way we use Grunt across all our services. We already use watch in some places I think, but I prefer to run it manually. I'm happy to accept a change like this, but I don't understand the use of 'express' rather than 'execute'.

dpasqualin:
'execute' runs the process on the foreground, while 'express' runs it on background, as required by 'watch'. The drawback is that on 'watch' mode the developer can't see the output generated by the 'app.js' application.

Perhaps we could use both, 'express' for `grunt watch` and 'execute' for `grunt default`?

PR #4, opened 2015-06-11
Replace UTF-16 surrogate characters in client updates

jpallen:


jpallen:
Closing this and moving the code to doc updater so that it also catches broken encoding via dropbox.

PR #6, opened 2016-10-24
Add end point to start draining clients

jpallen:


app/coffee/DrainManager.coffee
```diff
@@ -0,0 +1,25 @@
...
+	reconnectNClients: (io, N) ->
+		drainedCount = 0
+		for client in io.sockets.clients()
+			if !@RECONNECTED_CLIENTS[client.id]
+				@RECONNECTED_CLIENTS[client.id] = true
+				logger.log {client_id: client.id}, "Asking client to reconnect gracefully"
+				client.emit "reconnectGracefully"
+				drainedCount++
+			if drainedCount == N
+				break
```
henryoswald: a touch unclear about this break. Maybe me, i just get confused what will happen. 

could do something like this on line 17

maxClientsDrained = drainedCount == N
if !maxClientsDrained and !@RECONNECTED_CLIENTS[client.id]

jpallen: Thanks, changed.

PR #9, opened 2017-01-13
Return ranges from docupdater to client

jpallen:


app/coffee/WebsocketController.coffee
```diff
@@ -93,7 +93,7 @@ module.exports = WebsocketController =
 						escapedLines.push line
 					AuthorizationManager.addAccessToDoc client, doc_id
 					client.join(doc_id)
-					callback null, escapedLines, version, ops
+					callback null, escapedLines, version, ops, ranges
```
ShaneKilkelly: Function signature callback should be updated to indicate ranges will be provided to callback:

```
joinDoc: (client, doc_id, fromVersion = -1, callback = (error, doclines, version, ops, ranges) ->) ->
```

PR #12, opened 2017-05-02
Use new redis-sharelatex with support for cluster

jpallen:


briangough:
Need to update to redis-sharelatex v1.0.2, healthCheck method breaks

$ curl localhost:3026/health_check/redis
TypeError: Cannot call method 'nodes' of undefined

PR #13, opened 2017-05-11
Remove multi call to make compatible with redis-cluster

jpallen:


app/coffee/DocumentUpdaterManager.coffee
```diff
@@ -55,10 +55,8 @@ module.exports = DocumentUpdaterManager =
 	queueChange: (project_id, doc_id, change, callback = ()->)->
 		jsonChange = JSON.stringify change
 		doc_key = "#{project_id}:#{doc_id}"
-		multi = rclient.multi()
-		multi.rpush Keys.pendingUpdates({doc_id}), jsonChange
-		multi.sadd  "DocsWithPendingUpdates", doc_key
```
briangough: Is the DocsWithPendingUpdates set not used anywhere any more?

jpallen: No, it's got 4.5 million entries in the doc updater because real-time is adding to it, but doc updater does nothing with it.

briangough: Ouch!

briangough: :+1:

PR #18, opened 2017-09-21
Encoding changes & comments in ranges

40thieves:
Encodes comments & track changes as ASCII.

### DIFF
--- a/app/coffee/Router.coffee
+++ b/app/coffee/Router.coffee
@@ -100,7 +100,7 @@ module.exports = Router =
 					# Called with 4 args, things are as expected
 				else
 					logger.error { arguments: arguments }, "unexpected arguments"
-					return {} # ????
+					return # ????
 
 				WebsocketController.joinDoc client, doc_id, fromVersion, options, (err, args...) ->
 					if err?
--- a/app/coffee/Router.coffee
+++ b/app/coffee/Router.coffee
@@ -102,7 +102,7 @@ module.exports = Router =
 					logger.error { arguments: arguments }, "unexpected arguments"
 					return {} # ????
 
-				WebsocketController.joinDoc client, doc_id, options, fromVersion, (err, args...) ->
+				WebsocketController.joinDoc client, doc_id, fromVersion, options, (err, args...) ->
 					if err?
 						Router._handleError callback, err, client, "joinDoc", {doc_id, fromVersion}
 					else
--- a/app/coffee/Router.coffee
+++ b/app/coffee/Router.coffee
@@ -79,26 +79,23 @@ module.exports = Router =
 					if err?
 						Router._handleError null, err, client, "leaveProject"
 
-			# Variadic. The possible options:
+			# Variadic. The possible arguments:
 			# doc_id, callback
 			# doc_id, fromVersion, callback
 			# doc_id, options, callback
-			# doc_id, options, fromVersion, callback
-			client.on "joinDoc", (doc_id, options, fromVersion, callback) ->
-				# options is optional
+			# doc_id, fromVersion, options, callback
+			client.on "joinDoc", (doc_id, fromVersion, options, callback) ->
+				if typeof fromVersion == "function"
+					fromVersion = -1
+					options = {}
+					callback = fromVersion
+				else if typeof fromVersion == "object"
+					fromVersion = -1
+					options = fromVersion
+					callback = options
 				if typeof options == "function"
 					options = {}
 					callback = options
-					fromVersion = -1
-				else if typeof options == "number"
-					options = {}
-					fromVersion = options
-					callback = fromVersion
-
-				# fromVersion is optional
-				if typeof fromVersion == "function"
-					callback = fromVersion
-					fromVersion = -1
 
 				WebsocketController.joinDoc client, doc_id, options, fromVersion, (err, args...) ->
 					if err?
--- a/app/coffee/Router.coffee
+++ b/app/coffee/Router.coffee
@@ -79,8 +79,22 @@ module.exports = Router =
 					if err?
 						Router._handleError null, err, client, "leaveProject"
 
-
+			# Variadic. The possible options:
+			# doc_id, callback
+			# doc_id, fromVersion, callback
+			# doc_id, options, callback
+			# doc_id, options, fromVersion, callback
 			client.on "joinDoc", (doc_id, options, fromVersion, callback) ->
+				# options is optional
+				if typeof options == "function"
+					options = {}
+					callback = options
+					fromVersion = -1
+				else if typeof options == "number"
+					options = {}
+					fromVersion = options
+					callback = fromVersion
+
 				# fromVersion is optional
 				if typeof fromVersion == "function"
 					callback = fromVersion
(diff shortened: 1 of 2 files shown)


app/coffee/Router.coffee
```diff
@@ -80,13 +80,13 @@ module.exports = Router =
 						Router._handleError null, err, client, "leaveProject"
 
 
-			client.on "joinDoc", (doc_id, fromVersion, callback) ->
+			client.on "joinDoc", (doc_id, options, fromVersion, callback) ->
```
jpallen: We need to make this optional, since old clients will still be passing args with the old argument signature.

40thieves: Right of course. I'll fix that

app/coffee/WebsocketController.coffee
```diff
@@ -101,6 +101,29 @@ module.exports = WebsocketController =
...
+					if options.encodeRanges
+						if ranges.comments
+							escapedComments = []
+							for comment in ranges.comments
+								try
+									comment.op.c = unescape(encodeURIComponent(comment.op.c))
+								catch err
+									logger.err {err, project_id, doc_id, fromVersion, comment, client_id: client.id}, "error encoding comment uri component"
+									return callback(err)
+								escapedComments.push comment
```
jpallen: I don't think we need to collect these in a separate array. We're modifying in place above.

app/coffee/WebsocketController.coffee
```diff
@@ -101,6 +101,29 @@ module.exports = WebsocketController =
...
+							ranges.comments = escapedComments
+						if ranges.changes
+							escapedChanges = []
+							for change in ranges.changes
+								try
+									change.op.i = unescape(encodeURIComponent(change.op.i)) if change.op.i
```
jpallen: I think it's worth abstracting the `unescape(encodeURIComponent( ... ))` into an 'encodeForWebsockets` method now we use it in multiple places.

app/coffee/Router.coffee
jpallen: Thanks. I think `options` should be after `fromVersion`, since we may remove it after this is migrated. Also the end 'feels' like the right place for a flexible bundle of options (essentially named optional arguments).

40thieves:
Added some commits to make `options` optional and DRY up as you suggested.

There is a subtle different between what we log with 68b505a. After it we can't know which comment or change threw the error, so I just log out `ranges`

jpallen:
👍  I don't recall that encoding block ever throwing an error - it's just that it can, like `JSON.parse`, so we wrapped it for safety I think. I.e. it's not an expected error condition at all, so this compromise is ok IMO.

40thieves:
Added another commit which switches up the order of the variadic. I agree that it's better, I was just a little nervous about `typeof X == "object"` but I think it's fine

app/coffee/Router.coffee
```diff
@@ -79,14 +79,25 @@ module.exports = Router =
...
-
-			client.on "joinDoc", (doc_id, fromVersion, callback) ->
-				# fromVersion is optional
+			# Variadic. The possible arguments:
+			# doc_id, callback
+			# doc_id, fromVersion, callback
+			# doc_id, options, callback
+			# doc_id, fromVersion, options, callback
+			client.on "joinDoc", (doc_id, fromVersion, options, callback) ->
 				if typeof fromVersion == "function"
```
jpallen: Sorry to be nitpicky here... I'm finding the if statements below hard to reason about. Could we be really explicit here?

```
if typeof fromVersion == 'function' and !options?

else if typeof fromVersion == 'object' and typeof object == 'function'

else if typeof fromVersion == 'number' and typeof object == 'function'

else if typeof fromVersion == 'number' and typeof options == 'object'
   # pass, things are as expected
else
   logger.error "unexpect arguments"
   callback ( new Error(...) )
```

40thieves: No that's a good point. I had to think to reason it out

app/coffee/Router.coffee
```diff
@@ -79,14 +79,30 @@ module.exports = Router =
...
+					options = {}
+				else if typeof fromVersion == "object" and typeof options == "function"
+					callback = options
+					options = fromVersion
+					fromVersion = -1
+				else if typeof fromVersion == "number" and typeof options == "object"
+					# Called with 4 args, things are as expected
+				else
+					logger.error { arguments: arguments }, "unexpected arguments"
+					return # ????
```
40thieves: `callback` may not be defined here so we can rely on it to communicate the error back. So I wasn't sure what to do. WDYT?

jpallen: Hmm, good point. Do a best effort: `callback?(new Error(...))` (note the ? guard - my favourite CS feature that ES6 doesn't have!)

40thieves: Perhaps not for long https://github.com/tc39/proposal-optional-chaining

I'll fix that

jpallen: 😀  That would make me very happy!

app/coffee/WebsocketController.coffee
```diff
@@ -91,16 +91,29 @@ module.exports = WebsocketController =
...
 						catch err
 							logger.err {err, project_id, doc_id, fromVersion, line, client_id: client.id}, "error encoding line uri component"
 							return callback(err)
 						escapedLines.push line
+					if options.encodeRanges
+						try
+							for comment in ranges?.comments or []
+								comment.op.c = encodeForWebsockets(comment.op.c)
+							for change in ranges?.changes or []
+								change.op.i = encodeForWebsockets(change.op.i) if change.op.i
```
jpallen: Actually, one more little thing. We should use `change.op.i?` here. This is an explicit null/undefined check, where as `change.op.i` relies on JS's notion of 'truthy' which can catch us out on say an empty string being 'false'.

40thieves: Right good point. I'll fix

app/coffee/WebsocketController.coffee
```diff
@@ -91,16 +91,29 @@ module.exports = WebsocketController =
...
-							line = unescape(encodeURIComponent(line))
+							line = encodeForWebsockets(line)
 						catch err
 							logger.err {err, project_id, doc_id, fromVersion, line, client_id: client.id}, "error encoding line uri component"
 							return callback(err)
 						escapedLines.push line
+					if options.encodeRanges
+						try
+							for comment in ranges?.comments or []
+								comment.op.c = encodeForWebsockets(comment.op.c)
```
40thieves: I did actually mean to ask whether we need to be defensive here too?

jpallen: I think we can rely on this being present, but begin defensive would be consistent with the changes section below too, and generally a good practice. We don't want this crashing based on bad data/bad assumptions 👍

jpallen:
Just a note on deploy order: https://github.com/sharelatex/real-time-sharelatex/pull/18 will need to be deployed first, to ensure that the backend can support the new method signature first.

40thieves:
Note: I realised tests were failing after merging, so have https://github.com/sharelatex/real-time-sharelatex/pull/19/files in to fix before deploying

PR #26, opened 2017-12-12
If a user has only their `first_name` set, don't label as Anonymous, connects to sharelatex/web-sharelatex-internal/#96

ShaneKilkelly:


jpallen:
This looks like it could create `"undefined Allen"` if the user had a last name but no first name? Probably unlikely though.

ShaneKilkelly:
@jpallen, good catch. I think it's unlikely as we always populate the user's first name based on the username portion of their email address. So I'm pretty sure they'll always have a first name.

I may just fix that up though, especially with the integration work, we may not be able to rely on that implicit behaviour.

jpallen:
Do we also need to handle the 'no name' case and fall back to email (perhaps the bit before the @?)

ShaneKilkelly:
@jpallen In the no-name case they get set to "Anonymous", a bit further down in the code.

ShaneKilkelly:
@paulojreis I've updated to handle the last-name-only case, would you mind taking another look? :)

PR #30, opened 2018-12-04
Dockerised

henryoswald:
- Adds FORCE_DRAIN_MS_DELAY setting to cleanup connections after a sigterm
- Update build scripts to build docker and tar.gz file


app.coffee
```diff
@@ -73,3 +82,31 @@ server.listen port, host, (error) ->
...
+	logger.log {delay_ms:Settings.forceDrainMsDelay}, "starting force drain after timeout"
+	setTimeout ()-> 
+		logger.log "starting drain"
+		DrainManager.startDrain(io, 4)
+	, Settings.forceDrainMsDelay
+
+if Settings.forceDrainMsDelay?
+	Settings.forceDrainMsDelay = parseInt(Settings.forceDrainMsDelay, 10)
+	logger.log forceDrainMsDelay: Settings.forceDrainMsDelay,"forceDrainMsDelay enabled"
+	for signal in ['SIGINT', 'SIGHUP', 'SIGQUIT', 'SIGUSR1', 'SIGUSR2', 'SIGTERM', 'SIGABRT']
```
jpallen: I think we should limit this behaviour to a single signal, and then use that. Based on Nginx convention, `SIGQUIT` is for a 'graceful' shutdown. Or we may have our hands tied by the signal that k8s/docker will send? But we still want a way to terminate the process immediately if we need to I think.

jpallen: (sorry for the unrequested input!)

henryoswald: There are multiple signals which k8 will send this container depending on the reason for taking down the container. I know it sends `SIGABRT` for a delete pod request and a `SIGINT` for scale down requests, `SIGHUP` gets sent later on in the delete lifecycle I think. The above is just a catchall I took from doc updater to ensure I don't miss one.

jpallen: ok, cool, thanks, that makes sense 👍

app.coffee
```diff
@@ -73,3 +83,31 @@ server.listen port, host, (error) ->
...
+		process.on signal, ->
+			logger.log signal: signal, "received interrupt, cleaning up"
```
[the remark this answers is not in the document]
henryoswald: added https://github.com/sharelatex/real-time-sharelatex/commit/de081b18c23e5eb6580fcde0e1e2bc42c6ec6f9d WDYT?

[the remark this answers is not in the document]
briangough: Looks good.

PR #35, opened 2019-02-11
Nameless logged in user labelled as anonymous

chrystalgriffiths:
### Description

If a user joins in the middle of a user's session and they have neither first name nor last name set, they were labelled as Anonymous instead of by their email address as they should be.

Sends an empty string for the user's name instead of "Anonymous". Users who are truly anonymous are dealt with in web (See related PR).

#### Related Issues / PRs

Fixes https://github.com/sharelatex/web-sharelatex-internal/issues/910

Relates to https://github.com/sharelatex/web-sharelatex-internal/pull/1501

#### Manual Testing Performed

- [X] Truly anonymous user appears and is labelled as 'Anonymous'
- [X] Logged in user with no name appears and is labelled with their email address

### Deployment

Deploy the web one first https://github.com/sharelatex/web-sharelatex-internal/pull/1501

#### Who Needs to Know?

@jdleesmiller was pairing with me on it at one point so he will probably be interested in what happened with it in the end

### DIFF
--- a/app/coffee/WebsocketController.coffee
+++ b/app/coffee/WebsocketController.coffee
@@ -52,6 +52,10 @@ module.exports = WebsocketController =
 		metrics.inc "editor.leave-project"
 		Utils.getClientAttributes client, ["project_id", "user_id"], (error, {project_id, user_id}) ->
 			return callback(error) if error?
+
+			logger.log {project_id, user_id, client_id: client.id}, "client leaving project"
+			WebsocketLoadBalancer.emitToRoom project_id, "clientTracking.clientDisconnected", client.id
+			
 			# bail out if the client had not managed to authenticate or join
 			# the project.  Prevents downstream errors in docupdater from
 			# flushProjectToMongoAndDelete with null project_id.
@@ -61,9 +65,6 @@ module.exports = WebsocketController =
 			if not project_id?
 				logger.log {user_id: user_id, client_id: client.id}, "client leaving, not in project"
 				return callback()
-
-			logger.log {project_id, user_id, client_id: client.id}, "client leaving project"
-			WebsocketLoadBalancer.emitToRoom project_id, "clientTracking.clientDisconnected", client.id
 		
 			# We can do this in the background
 			ConnectedUsersManager.markUserAsDisconnected project_id, client.id, (err) ->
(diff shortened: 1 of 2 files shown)


40thieves:
> Sometimes the users would linger so that if I left and came back I would end up with even more duplication and 5 users showing.

Fyi, I've seen this on master/production. I think we're not always handling the disconnection well/immediately

40thieves:
Hmm, I now can't reproduce that problem. I previously thought it was an eventual consistency thing, since it seemed to fix itself over time, but now in my testing it appears fairly immediate.

jpallen:
I would like to see us get this code cleaned up so that `ConnectedUsersManager.updateUserPosition` is always called, or that we understand why it isn't called for anonymous users, and why doing so leaves these dangling users in the editor. Now that we've touched it, I'm concerned about just leaving a potentially unexploded mine for someone else in future. Shall we aim to pair on this sometime @chrystalgriffiths, and hopefully get to the bottom of it?

jdleesmiller:
Thanks for writing this up.

We had it mostly working without the `if first_name or last_name` that (I think) is intended to test for whether the user anonymous, but I guess that issue with the duplicated users came back after we finished. Do you think that's the intention of the `if first_name or last_name`, or is it something else, @jpallen ?

I ventured a possible explanation: there can be a lot of anonymous users on a read only view, so calling `updateClientPosition` might create a lot of them and cause some performance / display problem. However, if we're handling that well elsewhere, that may be a non issue.

chrystalgriffiths:
I think pairing and getting to the bottom of it is a good idea @jpallen 

It would be nice to get it working with `updateUserPosition`.

jpallen:
> I ventured a possible explanation: there can be a lot of anonymous users on a read only view, so calling `updateClientPosition` might create a lot of them and cause some performance / display problem. However, if we're handling that well elsewhere, that may be a non issue.

This is my hunch as to why we did it this way in the first place. But the test we're using at the moment isn't actually explicitly picking out anonymous users. And I don't see why it's causing the dangling users in the editor after disconnect. So *something* is happening here that I think we should understand.

chrystalgriffiths:
Just a general FYI for anyone reading this - the multiple users was just because of Redis not clearing itself properly when you restart the real-time container.

We've changed it so that we still don't send cursor data up for truly anonymous users just because that's how it was before and maybe we will get an influx of anonymous users on some project one time and we don't want to affect performance.

Nameless logged in users will send their data up.

PR #37, opened 2019-02-15
setup continualPubsubTraffic to keep pub sub happy

henryoswald:
### Description
this keeps the pub sub channel ticking along happily. Likely only an issue in staging but it means there is always traffic happening. I think the root cause is ghostunnel but this solves the issue to an adequate degree. 

#### Related Issues / PRs
https://github.com/sharelatex/overleaf-google-ops/issues/176

#### Potential Impact
Increases the amount of traffic through redis, but even with 30 instances it is only an extra 0.5ms which is nothing for redis.

#### Manual Testing Performed
updates on https://staging-google.sharelatex.com should keep working and not timeout

### DIFF
--- a/app.coffee
+++ b/app.coffee
@@ -1,6 +1,7 @@
 Metrics = require("metrics-sharelatex")
 Settings = require "settings-sharelatex"
 Metrics.initialize(Settings.appName or "real-time")
+async = require("async")
 
 logger = require "logger-sharelatex"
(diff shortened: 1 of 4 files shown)


henryoswald:
The Docupdater will be publishing to `applied-ops`, it doesn't subscribe to it. `editor-events` is isolated to real time.

PR #38, opened 2019-03-19
add check for duplicate events

briangough:
### Description

Check message ids on pubsub message events to detect duplicate and out-of-order events.  

#### Screenshots

NA

#### Related Issues / PRs

Connects to https://github.com/overleaf/sharelatex/issues/1627 and https://github.com/sharelatex/web-sharelatex-internal/pull/1639

### Review

Adds a `EventLogger` module with a `checkEventOrder` method, keeps track of counters for each host key.  Clears out stale host keys every 100000 events, deleting any which have not received updates in the past hour.

#### Potential Impact

Low (as long as it doesn't crash) - doesn't modify messages.

#### Manual Testing Performed

- [X] Unit tests
- [X] Inject errors in dev environment

#### Accessibility

NA

### Deployment

No special issues, will work whether message id is present or not.

#### Deployment Checklist

- [ ]  Check chat messages are working

#### Metrics and Monitoring

Errors will be logged in sentry

#### Who Needs to Know?

@henryoswald

### DIFF
--- a/app/coffee/EventLogger.coffee
+++ b/app/coffee/EventLogger.coffee
@@ -1,4 +1,5 @@
 logger = require 'logger-sharelatex'
+metrics = require 'metrics-sharelatex'
 
 # keep track of message counters to detect duplicate and out of order events
 # messsage ids have the format "UNIQUEHOSTKEY-COUNTER"
@@ -11,7 +12,7 @@ module.exports = EventLogger =
 
 	MAX_STALE_TIME_IN_MS: 3600 * 1000
 
-	checkEventOrder: (message_id, message) ->
+	checkEventOrder: (channel, message_id, message) ->
 		return if typeof(message_id) isnt 'string'
 		[key, count] = message_id.split("-", 2)
 		count = parseInt(count, 10)
@@ -20,12 +21,15 @@ module.exports = EventLogger =
 		# store the last count in a hash for each host
 		previous = EventLogger._storeEventCount(key, count)
 		if !previous? || count == (previous + 1)
+			metrics.inc "event.#{channel}.valid"
 			return # order is ok
 		if (count == previous)
-			logger.error {key:key, previous: previous, count:count, message:message}, "duplicate event"
+			metrics.inc "event.#{channel}.duplicate"
+			# logger.error {key:key, previous: previous, count:count, message:message}, "duplicate event"
 			return "duplicate"
 		else
-			logger.error {key:key, previous: previous, count:count, message:message}, "events out of order"
+			metrics.inc "event.#{channel}.out-of-order"
+			# logger.error {key:key, previous: previous, count:count, message:message}, "events out of order"
 			return # out of order
 
 	_storeEventCount: (key, count) ->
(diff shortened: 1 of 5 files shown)


app/coffee/EventLogger.coffee
```diff
@@ -0,0 +1,46 @@
...
+		return if typeof(message_id) isnt 'string'
+		[key, count] = message_id.split("-", 2)
+		count = parseInt(count, 10)
+		if !count # ignore checks if counter is not present
+			return
+		# store the last count in a hash for each host
+		previous = EventLogger._storeEventCount(key, count)
+		if !previous? || count == (previous + 1)
+			return # order is ok
+		if (count == previous)
```
jpallen: I think this could be `<=`, since it's possible it's duplicate, and out of order?

jpallen: Actually, I see the else branch covers that, so ignore.

app/coffee/EventLogger.coffee
```diff
@@ -0,0 +1,46 @@
...
+		now = Date.now()
+		EVENT_LOG_COUNTER[key] = count
+		EVENT_LOG_TIMESTAMP[key] = now
+		# periodically remove old counts
+		if (++EVENT_COUNT % EventLogger.MAX_EVENTS_BEFORE_CLEAN) == 0
+			EventLogger._cleanEventStream(now)
+		return previous
+
+	_cleanEventStream: (now) ->
+		for key, timestamp of EVENT_LOG_TIMESTAMP
```
jpallen: Just to double check my understanding, the `key` is per web process, so we're looking on the order of ~25 keys per deploy/restart of web, so this isn't going to grow to be an expensive operation?

jpallen: And the reason for cleaning is to get rid of keys that correspond to no-longer live versions of web?

briangough:
Thanks.  I agree with adding metrics.

What do you think about running EventLogger on the DocumentUpdaterController `applied-ops` channel by getting docupdater to send a unique counter like web?  https://github.com/sharelatex/real-time-sharelatex/blob/401dd06945565482f26b29f49ccb0770bac00b82/app/coffee/DocumentUpdaterController.coffee#L13-L30

jpallen:
> What do you think about running EventLogger on the DocumentUpdaterController applied-ops channel by getting docupdater to send a unique counter like web?

Yes, that would be what we'd need to do to monitor/debug the 'out of sync' problems further, so sounds good. Are you happy to add this in?

app/coffee/EventLogger.coffee
briangough: Correct, one key per web process, if there are no updates for an hour it is removed.  Since we don't deploy realtime very often I thought we should clean them out (to be honest the memory usage should be minimal even if it was running for months).  I've got it cleaning every 100,000 operations but maybe making it time-based, e.g. once per day, is simpler.

briangough:
Yes, I will make a PR for docupdater.

PR #40, opened 2019-04-09
handle duplicate entries in io.sockets.clients

briangough:
### Description

We think this is the root cause of out of sync errors.  The socket.io library can get into a state where there are duplicate entries when iterating over the clients in a 'room', causing duplicate messages to be sent (https://github.com/overleaf/sharelatex/issues/1700, https://github.com/overleaf/sharelatex/issues/1632 and https://github.com/overleaf/sharelatex/issues/1627)

#### Screenshots

NA

#### Related Issues / PRs

https://github.com/overleaf/sharelatex/issues/1700

### Review

Keeps the same iteration but uses a `seen[id]` hash to skip sending messages to clients already seen.

#### Potential Impact

High

#### Manual Testing Performed

- [x] Unit and acceptance tests  (extended the unit tests to fail if duplicates are not removed)
- [x] Local dev env

#### Accessibility

NA

### Deployment

NA

#### Deployment Checklist

NA

#### Metrics and Monitoring

Grafana realtime dashboard, new metric available `socket-io.duplicate-clients`

#### Who Needs to Know?

@jpallen @jdleesmiller

### DIFF
--- a/app/coffee/DocumentUpdaterController.coffee
+++ b/app/coffee/DocumentUpdaterController.coffee
@@ -4,6 +4,7 @@ redis = require("redis-sharelatex")
 rclient = redis.createClient(settings.redis.documentupdater)
 SafeJsonParse = require "./SafeJsonParse"
 EventLogger = require "./EventLogger"
+metrics = require "metrics-sharelatex"
 
 MESSAGE_SIZE_LOG_LIMIT = 1024 * 1024 # 1Mb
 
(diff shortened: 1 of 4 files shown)


briangough:
For the record this did not work.  There are no duplicate clients, the pubsub subscribe callback was being called multiple times for each event, which looked the same in the logs.  This PR could be reverted as some point.

PR #44, opened 2019-04-15
add health check to pubsub channels

briangough:
### Description

This adds a health check to the realtime `applied-ops` and `editor-events` pubsub channels, piggybacking on the existing `continualPubSubTraffic` code.  

The health check adds a unique id to each continual traffic message and checks that it is received once (only) within a timeout of 3 seconds.    If we begin to see duplicate messages (as in overleaf/sharelatex#1700) or no messages then we should restart real-time. 

It also monitors the latency of the pubsub channel as we do not currently track this.  

I have increased the frequency of the continual pubsub traffic to one message every 20 seconds, to match the default health check frequency. 

#### Screenshots

NA

#### Related Issues / PRs

overleaf/sharelatex#1700

### Review

Self-contained change mainly in `HealthCheckManager`.

#### Potential Impact

Low

#### Manual Testing Performed

- [X]  Tested in dev env with simulated errors

#### Accessibility

NA

### Deployment

NA

#### Deployment Checklist

- [ ] Check latency graphs

#### Metrics and Monitoring

Grafana metrics for `stats.timer....event.channel.latency`

### DIFF
--- a/app.coffee
+++ b/app.coffee
@@ -20,6 +20,7 @@ SessionSockets = require('session.socket.io')
 CookieParser = require("cookie-parser")
 
 DrainManager = require("./app/js/DrainManager")
+HealthCheckManager = require("./app/js/HealthCheckManager")
 
 # Set up socket.io server
 app = express()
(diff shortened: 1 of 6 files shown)


jdleesmiller:
Thought I would have a look out of interest... I don't know much about how this works, but is there a concern about leaking instances of this class? I can see them being added to `CHANNEL_MANAGER` but never deleted. Is the number of channels bounded?

briangough:
There are only 2 channels so they should get garbage collected after the new one overwrites the entry for the old one.

jdleesmiller:
> There are only 2 channels

Great, thanks. Now I know.

PR #45, opened 2019-04-15
update request module

briangough:
### Description

Fixes vulnerability reported by github: `Upgrade request to version 2.68.0 or later.`

> CVE-2017-16026  moderate severity
> Vulnerable versions: >= 2.2.6, < 2.47.0
> Patched version: 2.68.0
> Affected versions of request will disclose local system memory to remote systems in certain circumstances. When a multipart request is made, and the type of body is number, then a buffer of that size will be allocated and sent to the remote server as the body.

#### Screenshots

NA

#### Related Issues / PRs

https://github.com/sharelatex/real-time-sharelatex/network/alert/package.json/request/open

### Review

Updates request module only

#### Potential Impact

High but should be safe.

#### Manual Testing Performed

- [X] Acceptance tests

#### Accessibility

NA

### Deployment

NA

#### Deployment Checklist

NA

#### Metrics and Monitoring

NA

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "logger-sharelatex": "^1.6.0",
     "metrics-sharelatex": "^2.1.1",
     "redis-sharelatex": "git+https://github.com/sharelatex/redis-sharelatex.git#v1.0.6",
-    "request": "~2.34.0",
+    "request": "^2.34.0",
     "session.socket.io": "^0.1.6",
     "settings-sharelatex": "^1.1.0",
     "socket.io": "0.9.16",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -31,7 +31,7 @@
     "logger-sharelatex": "^1.6.0",
     "metrics-sharelatex": "^2.1.1",
     "redis-sharelatex": "git+https://github.com/sharelatex/redis-sharelatex.git#v1.0.6",
-    "request": "~2.34.0",
+    "request": "^2.34.0",
```
henryoswald: should this not read `2.88.0`?

briangough: ^2.34 will update to the latest before 3.0, so it actually gets 2.88 (which is the latest)
(~2.34.0 only updated to 2.34.x)
I can change it if it's clearer.

henryoswald: I think we should change it as it is quite a few versions behind and people often look at the package.json to see what is installed.

briangough: :+1: done

PR #53, opened 2019-05-31
Do not log error on doc updater 404

emcsween:
### Description

Document updater 404s are not indicative of a problem, but just of the client
trying to connect to a deleted document, which is easy to trigger. We
log a warning instead.

I also changed the assertions for errors because comparing errors is a tricky business. In our case, tests would pass even if the error messages were not those expected.

#### Related Issues / PRs

This solves overleaf/issues#1788

#### Manual Testing Performed

One way I found to reproduce the problem:

- [x] Look at the real-time logs
- [x] Open a document
- [x] While in the document, delete it
- [x] Continue working on the document until Overleaf complains it's out of sync
- [x] Verify that real-time didn't log anything at ERROR level


briangough:
Will deploy this as I am deploying some other pull requests for real-time this morning.

PR #58, opened 2019-07-01
support multple redis instances for pubsub

briangough:
### Description

Allow multiple redis configurations for pubsub by replacing the single `rclient`s used for pubsub by a list `rclientList`. 

The `applied-ops` and `editor-events` channels on both redis configurations are subscribed to, the services creating events on those (docupdater and web) can be switched over gracefully.

#### Screenshots

NA

#### Related Issues / PRs

https://github.com/overleaf/google-ops/issues/333

### Review

Health checks not extended yet TBD
Needs testing locally

#### Manual Testing Performed

- [X] Unit and acceptance tests still work for single redis

#### Accessibility

NA

### Deployment

Deploy this in advance and move everyone over to this version - then other services can be switched when needed. 

#### Deployment Checklist

- [ ] Add settings for second redis configuration
- [ ] Deploy this version
- [ ] Drain old version or wait for everyone to move over to this version
- [ ] Switch one web server to use `editor-events` on second redis
- [ ] If ok, switch all web servers to second redis
- [ ] Switch one docupdater to use `applied-ops` on second redis
- [ ] If ok, switch all docupdaters to second redis
- [ ] Cleanup, remove old client from list so only the new one is in use.

#### Metrics and Monitoring

Docupdater update rate

#### Who Needs to Know?

@henryoswald @mans0954 @jdleesmiller

### DIFF
--- a/app/coffee/DocumentUpdaterController.coffee
+++ b/app/coffee/DocumentUpdaterController.coffee
@@ -1,7 +1,6 @@
 logger = require "logger-sharelatex"
 settings = require 'settings-sharelatex'
 redis = require("redis-sharelatex")
-rclient = redis.createClient(settings.redis.pubsub)
 SafeJsonParse = require "./SafeJsonParse"
 EventLogger = require "./EventLogger"
 HealthCheckManager = require "./HealthCheckManager"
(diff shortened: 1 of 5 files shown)


henryoswald:
It should be ok as it is summer but this will double the load on the real time processes.

briangough:
I think the total event rate (and load) should still be the same, but some events will be coming from a different redis when we switch over docupdater/web.  In docupdater and web each event will only be published to one redis.

henryoswald:
FYI i've been adding the `pubsub` redis connection in to the apps. https://github.com/overleaf/real-time/pull/59/files#diff-00b885097b3922589267200f7a2729eeR4

briangough:
I'll rebase this onto https://github.com/overleaf/real-time/pull/59 when it is merged.

henryoswald:
I guess the final deployment step in your checklist would be to remove `redis.createClient(Settings.redis.documentupdater)` from the client list?

briangough:
Yes, afterwards we can remove the old client from the list.

app/coffee/DocumentUpdaterController.coffee
```diff
@@ -12,13 +11,15 @@ MESSAGE_SIZE_LOG_LIMIT = 1024 * 1024 # 1Mb
 module.exports = DocumentUpdaterController =
 	# DocumentUpdaterController is responsible for updates that come via Redis
 	# Pub/Sub from the document updater.
+	rclientList: [redis.createClient(settings.redis.pubsub)]
```
henryoswald: were you thinking get the multi client setup live with a single client and then when we are ready todo the migration add `redis.createClient(settings.redis.pubsub)` in a second deploy?

briangough: yes, this puts the framework in place, then we can add extra entries to the array.

briangough: if we know what we want, we can add it now.

briangough:
I just made three changes.. including a fix for a mistake.

1.  there was a bug in WebsocketLoadbalancer.  It was publishing to both the old and new redis clients, and listening to them as well.  This was a mistake. We should only publish to one, but listen to both.  The final migration happens when we switch which one we **publish** to.
2.  I changed the code so that it creates the list of clients dynamically.  If a client isn't defined it doesn't add it to the list.  So we can then do everything by changing the configuration only.
3. I added a per-client metric for `applied-ops` so we can see how much traffic is going through each client.

henryoswald:
Great. I have already uploaded the unusedpubsub into chef btw.

PR #61, opened 2019-07-11
remove redis cluster from realtime

briangough:
### Description

Remove the additional `unused` redis client for cluster, so that only the primary `pubsub` redis sentinel client is used.

Add a check to skip the per-client metrics when there is only one client.

Retaining the infrastructure to support multiple clients in case we need to do any further migrations. 

#### Screenshots

NA

#### Related Issues / PRs

https://github.com/overleaf/google-ops/issues/333

### Review

Small change.

#### Potential Impact

Low

#### Manual Testing Performed

- [ ] Unit and acceptance tests

#### Accessibility

NA

### Deployment

NA

#### Deployment Checklist

- [ ]  Check traffic on `SUBSCRIBE applied-ops` and `SUBSCRIBE editor-events` on redis cluster.  There should be none.

#### Metrics and Monitoring

NA


briangough:
For deployment next week.

PR #65, opened 2019-07-18
listen on separate channels for each project/doc

briangough:
### Description

This PR adds support for individual project and doc pubsub channels, `editor-events:#{project_id}` and `applied-ops:#{doc_id}`.  It still listens to the main `editor-events` and `applied-ops` channels and handles messages in the same way, whether they arrive on the main or individual channels.

The idea is to reduce the load on the redis server, because it now only needs to send events to real-time processes that have socket.io connections for a specific project/doc.   Currently all the events are sent to every real-time process, and real-time checks to see if it has any socket.io clients for the relevant project/doc - discarding the event if it does not.  Currently we have 30x live realtime processes and 30x dark realtime processes, which makes a 60x amplification of each event published from docupdater or web.  Typically projects have 1 user connected to a single real-time process, so this is very wasteful.

Note: realtime itself publishes on the `editor-events` channel, so this traffic cannot be sent to the individual channels until all processes (dark and live) are listening on the individual channels.  The setting `publishOnIndividualChannels` controls this, it should only be set to `true` when all the clients are on the new deploy.

There are now unit tests for ChannelManager and RoomManager.

#### Screenshots

NA

#### Related Issues / PRs

https://github.com/overleaf/document-updater/pull/77
https://github.com/overleaf/web-internal/pull/2003

#### Rollback

The main safety mechanism we have is that the existing code path for the main `editor-events` and `applied-ops` channels is not changed at all and continues to operate alongside the new channels.  As long as we can get back to events being sent to the main `editor-events` and `applied-ops`  channels from web and docupdater we can roll back to a safe state.

#### Naming

I have separated the socket.io code into RoomManager (socket.io refers to clients being in `rooms`) and the redis pubsub logic into   ChannelManager (redis refers to pubsub `channels`).

####  Subscribe/unsubscribe logic

We want to be able to unsubscribe from a channel when it is no longer needed, to minimise redis traffic.  We should never subscribe more than once to a channel, that could create problems in the driver (we've already be hit by this before).

I have chosen is to create `project-active` and `project-empty` events which fire on the first client to enter and last client to leave the room.  These are used to trigger corresponding subscribe and unsubscribe actions on the redis client.  This way we only subscribe on the first client to join the room.  

We have to check the socket.io internal state for this, we already do this for `_getConnectedClientView` in  [HttpController.coffee#L5-L22](https://github.com/overleaf/real-time/blob/605c4afae6bb728ea5d1f0c197c3bb5c21ccbae8/app/coffee/HttpController.coffee#L5-L22)

All of the socket.io code for joining and leaving rooms is synchronous, so we will get consistent results for the number of client in a room within our own methods.  Note that the redis subscribe/unsubscribe calls are asynchronous. To avoid missing events when we join a doc (or project), we capture the promise from the subscribe and only add the client to the room and send the doc to the client when it has resolved.  We save the promise in a map for each client so that subsequent joins can reuse any pending  subscribe. 

We don't need to do this for when we unsubscribe from a per-doc or per-project channel, if there are events which come in after the client has left the room then it is harmless.
 
#### Potential Impact

High, core editor functionality.   Deployment in stages should make it possible to roll back if problems are detected.

#### Manual Testing Performed

- [x] Unit and acceptance tests
- [x] Tested in local dev env

#### Accessibility

NA

### Deployment

Migration plan:
- [ ]  deploy real-time and drain dark to get all clients listening to individual channels
- [ ] change web to publish on `editor-events:#{project_id}` channel (https://github.com/overleaf/web-internal/pull/2003)
- [ ] change 1 docupdater to publish on individual `applied-ops:#{doc_id}` channels (https://github.com/overleaf/document-updater/pull/77)
- [ ] change all docupdaters to do this
- [ ] deploy realtime with setting `publishOnIndividualChannels:true`

Remember to update relevant settings in google-ops and chef

#### Metrics and Monitoring

New metrics `room-listeners`, `subscribe.applied-ops`, `subscribe.editor-events`.

#### Who Needs to Know?

@henryoswald


app/coffee/ChannelManager.coffee
```diff
@@ -0,0 +1,44 @@
...
+
+module.exports = ChannelManager =
+    _createNewClientEntry: (rclient) ->
+        ClientMap.set(rclient, new Set()).get(rclient)
+
+    subscribe: (rclient, baseChannel, id) ->
+        existingChannelSet = ClientMap.get(rclient) || @_createNewClientEntry(rclient)
+        channel = "#{baseChannel}:#{id}"
+        if existingChannelSet.has(channel)
+            logger.error {channel}, "already subscribed"
```
henryoswald: is this an error?

app/coffee/ChannelManager.coffee
```diff
@@ -0,0 +1,44 @@
...
+        else
+            rclient.subscribe channel
```
das7pad: Possible race condition here. Consider adding the channel to the `existingChannelSet` first.

app/coffee/ChannelManager.coffee
```diff
@@ -0,0 +1,44 @@
...
+            metrics.inc "unsubscribe.#{baseChannel}"
+
+    publish: (rclient, baseChannel, id, data) ->
+        if id is 'all' or !settings.publishOnIndividualChannels
+            channel = baseChannel
+        else
+            channel = "#{baseChannel}:#{id}"
+        # we publish on a different client to the subscribe, so we can't
+        # check for the channel existing here
+        rclient.publish channel, data
```
das7pad: Missing trailing new line.

app/coffee/RoomManager.coffee
```diff
@@ -0,0 +1,69 @@
...
+    _leave: (client, entity, id) ->
+        beforeCount = @_clientsInRoom(client, id)
+        client.leave id
+        afterCount = @_clientsInRoom(client, id)
+        logger.log {client: client.id, entity, id, beforeCount, afterCount}, "client left room"
+        # is the room now empty? if so, unsubscribe
+        if beforeCount == 1 and afterCount == 0
+            logger.log {entity, id}, "room is now empty"
+            RoomEvents.emit "#{entity}-empty", id
+            IdMap.delete(id)
```
das7pad: Missing trailing new line.

app/coffee/ChannelManager.coffee
briangough: Thanks, I forgot the `subscribe` is asynchronous.

app/coffee/ChannelManager.coffee
briangough: now downgraded to a warning

briangough:
I've updated the code to ensure that the subscribe has completed before the client joins the doc/project, so that we don't miss events that come in as the client is joining.

app/coffee/ChannelManager.coffee
```diff
@@ -0,0 +1,57 @@
+logger = require 'logger-sharelatex'
+metrics = require "metrics-sharelatex"
+settings = require "settings-sharelatex"
+
+ClientMap = new Map() # for each redis client, store a Map of subscribed channels (channelname -> subscribe promise)
```
henryoswald: To confirm I think these rclients are just to support multiple backends, not a redis client for every subscribe?

briangough: Correct.

app/coffee/DocumentUpdaterController.coffee
```diff
@@ -27,7 +29,18 @@ module.exports = DocumentUpdaterController =
...
 						metrics.inc "rclient-#{i}", 0.001 # per client event rate metric
-		
+		@handleRoomUpdates(@rclientList)
+
+	handleRoomUpdates: (rclientSubList) ->
+		roomEvents = RoomManager.eventSource()
+		roomEvents.on 'doc-active', (doc_id) ->
+			subscribePromises = for rclient in rclientSubList
+				ChannelManager.subscribe rclient, "applied-ops", doc_id
+			RoomManager.emitOnCompletion(subscribePromises, "doc-subscribed-#{doc_id}")
```
henryoswald: the listeners for this are not jumping out at me.

app/coffee/RoomManager.coffee
```diff
@@ -0,0 +1,93 @@
...
+    eventSource: () ->
+        return RoomEvents
+
+    joinEntity: (client, entity, id, callback) ->
+        beforeCount = @_clientsInRoom(client, id)
+        # is this a new room? if so, subscribe
+        if beforeCount == 0
+            logger.log {entity, id}, "room is now active"
+            RoomEvents.once "#{entity}-subscribed-#{id}", (err) ->
+                # only allow the client to join when all the relevant channels have subscribed
```
henryoswald: If there are ops processed between the socket.io connection and the joining of the room does the client need to reconnect?

briangough: Hmm. Good point. 

I changed the ordering at [WebsocketController.coffee#L95-L99](https://github.com/overleaf/real-time/blob/4c81e407d2fd3bf158dfb310c82a6268d3cf1c27/app/coffee/WebsocketController.coffee#L95-L99) so that the channel is subscribed and client goes into the room before we request the doc from docupdater and send it to them - this ensures that they don't miss any ops that have been processed after the doc is requested from docupdater.   But it does mean the client could be sent an applied-op before they have the doc and I had overlooked that.  I have just checked in web and this is handled ok, the web frontend to will discard it.

We bind to the socket events early  at [Document.js#L102](https://github.com/overleaf/web-internal/blob/b52fcdd551e6fe0a75a4b051237a9d5f483deaf9/public/src/ide/editor/Document.js#L102), before even joining the doc.  Updates arriving before the doc is joined are handled because the front end only accept updates when `doc` is not null at  [Document.js#L422](https://github.com/overleaf/web-internal/blob/b52fcdd551e6fe0a75a4b051237a9d5f483deaf9/public/src/ide/editor/Document.js#L422), so nothing gets applied until the doc is fully  joined.  So we are ok on this one.

PR #69, opened 2019-08-13
add shutdownDrainTimeWindow, drains all connections within time range

henryoswald:
### Description
Allows us to specify how long a process has to get everyone off the process.

#### Related Issues / PRs
https://github.com/overleaf/google-ops/issues/438

#### Manual Testing Performed

- [ ]
- [ ]

### DIFF
--- a/app.coffee
+++ b/app.coffee
@@ -110,17 +110,10 @@ shutdownCleanly = (signal) ->
 			shutdownCleanly(signal)
 		, 10000
 
-forceDrain = ->
-	logger.log {delay_ms:Settings.forceDrainMsDelay}, "starting force drain after timeout"
-	setTimeout ()-> 
-		logger.log "starting drain over #{Settings.shutdownDrainTimeWindow} mins"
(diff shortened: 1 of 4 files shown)


henryoswald:
I don't know if I like this anymore. Going to hold off merging while I think. 

I'm worried it will cause too much jumping between pods as they are deleted. I suppose we need to really trust the reconnect be be fast

briangough:
I think we need this PR anyway, because realtime pods will get shut down by kubernetes and we need to get users off them in time... whether we want to enforce the finite lifetime is a different matter.

henryoswald:
I actually said this on the wrong PR, I meant I wasn't sure about https://github.com/overleaf/real-time/pull/71/files

henryoswald:
i've removed forceDrainMsDelay, we should have just 1 flow for pods being spun down, either from deploys, auto scale down or nodes being killed. I am also going to bring the grace period for real time down to 10 mins in line with GKE node tear down times.

briangough:
:+1:

PR #73, opened 2019-08-14
add connected client count metric

briangough:
### Description

Record the number of connected clients for each real-time process

#### Screenshots

NA

### Review

Copied from DrainManager to get the number of connected clients

#### Potential Impact

Low, metric only

#### Manual Testing Performed

- [X] Unit and acceptance tests

#### Accessibility

NA

### Deployment

NA

#### Deployment Checklist

NA

#### Metrics and Monitoring

`socket-io.clients`

### DIFF
--- a/app/coffee/Router.coffee
+++ b/app/coffee/Router.coffee
@@ -78,7 +78,7 @@ module.exports = Router =
 
 			client.on "disconnect", () ->
 				metrics.inc('socket-io.disconnect')
-				metrics.gauge('socket-io.clients', io.sockets.clients()?.length)
+				metrics.gauge('socket-io.clients', io.sockets.clients()?.length - 1)
 				WebsocketController.leaveProject io, client, (err) ->
 					if err?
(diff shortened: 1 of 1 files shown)


briangough:
I've used them in project history and didn't have any problems.

app/coffee/Router.coffee
```diff
@@ -77,6 +78,7 @@ module.exports = Router =
 
 			client.on "disconnect", () ->
 				metrics.inc('socket-io.disconnect')
+				metrics.gauge('socket-io.clients', io.sockets.clients()?.length - 1)
```
das7pad: I would rather place it after the leaveProject call.

Consider a case with two connected clients. Now both are disconnecting at the same time. The metric will likely remain at 1.

jdleesmiller:
@das7pad Thanks --- we're only running one instance per host in the environment where we're using statsd (we are currently migrating to prometheus), so I think it will be OK.

PR #76, opened 2019-08-16
use a separate field to improve the connected client logic

briangough:
### Description

Fix logic for clients that have expired from redis.  If the client record expired, the `last_updated_at` field would be added back on its own when we ping the clients.  This made it look like the user was connected, but their record was empty.  

The fix for this is to check that the user_id field is present so that undefined users never show up in the list. 

#### Screenshots

NA

#### Related Issues / PRs

https://github.com/overleaf/issues/issues/2179

### Review

The additional check for `!result.user_id` fixes the problem.

I've also moved the connection tracking timestamp to a separate `last_response_at` field (the last time we successfully checked the client was connected) to avoid confusion with the `last_updated_at` field (the last time the user moved the cursor).

#### Potential Impact

Low

#### Manual Testing Performed

- [ ] Unit and acceptance tests

#### Accessibility

NA

### Deployment

NA

#### Deployment Checklist

NA

#### Metrics and Monitoring

NA

### DIFF
--- a/app/coffee/ConnectedUsersManager.coffee
+++ b/app/coffee/ConnectedUsersManager.coffee
@@ -25,6 +25,7 @@ module.exports =
 		multi.expire Keys.clientsInProject({project_id}), FOUR_DAYS_IN_S
 		
 		multi.hset Keys.connectedUser({project_id, client_id}), "last_updated_at", Date.now()
+		multi.hset Keys.connectedUser({project_id, client_id}), "socket_checked_at", Date.now()
 		multi.hset Keys.connectedUser({project_id, client_id}), "user_id", user._id
 		multi.hset Keys.connectedUser({project_id, client_id}), "first_name", user.first_name or ""
 		multi.hset Keys.connectedUser({project_id, client_id}), "last_name", user.last_name or ""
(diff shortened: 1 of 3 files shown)


app/coffee/ConnectedUsersManager.coffee
```diff
@@ -42,7 +43,7 @@ module.exports =
 	refreshClient: (project_id, client_id, callback = (err) ->) ->
 		logger.log project_id:project_id, client_id:client_id, "refreshing connected client"
 		multi = rclient.multi()
-		multi.hset Keys.connectedUser({project_id, client_id}), "last_updated_at", Date.now()
+		multi.hset Keys.connectedUser({project_id, client_id}), "socket_checked_at", Date.now()
```
das7pad: You could expire the `last_response_at` key directly after `REFRESH_TIMEOUT_IN_S` and save some processing overhead.

Namely the date creation here, and most of the parsing and date logic below.

`_isInactive` is then as simple as `return connected_user.socket_checked_at? `.

I would rename the key once again to `is_active`.

I was about to suggest this in the last PR but refused to due to the storage overhead in redis.

briangough:
I've moving this to the backlog in favour of deploying the simpler fix in https://github.com/overleaf/real-time/pull/77 today.

briangough:
Closing this in favour of using the Redis TTL in future as suggested by @das7pad  in https://github.com/overleaf/real-time/pull/76#discussion_r314639850.

PR #78, opened 2019-08-31
Generate retryable error when hitting rate limits in web

gh2k:
### Description

As part of the testing for overleaf/issues#2176, we discovered that it's easy to hit the project-join rate limit when reconnecting lots of times. When that happens, the connection fails pretty hard and you see a big scary warning.

This patch flags 429 as a retryable error, so that we can add logic to the client to handle this more gracefully.

#### Related Issues / PRs

overleaf/issues#2176

### Review

Added an `Errors` mechanism so that we can (in the future) generate multiple types of retryable error.

#### Who Needs to Know?

cc @timothee-alby, @henryoswald

### DIFF
--- a/app/coffee/Errors.coffee
+++ b/app/coffee/Errors.coffee
@@ -1,0 +1,10 @@
+CodedError = (message, code) ->
+	error = new Error(message)
+	error.name = "CodedError"
+	error.code = code
+	error.__proto__ = CodedError.prototype
+	return error
+CodedError.prototype.__proto__ = Error.prototype
+
+module.exports = Errors =
+	CodedError: CodedError
--- a/app/coffee/WebApiManager.coffee
+++ b/app/coffee/WebApiManager.coffee
@@ -1,6 +1,7 @@
 request = require "request"
 settings = require "settings-sharelatex"
 logger = require "logger-sharelatex"
+{ CodedError } = require "./Errors"
 
 module.exports = WebApiManager =
 	joinProject: (project_id, user, callback = (error, project, privilegeLevel) ->) ->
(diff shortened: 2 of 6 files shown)


app/coffee/WebApiManager.coffee
```diff
@@ -25,6 +26,10 @@ module.exports = WebApiManager =
 			if 200 <= response.statusCode < 300
 				callback null, data?.project, data?.privilegeLevel
 			else
-				err = new Error("non-success status code from web: #{response.statusCode}")
-				logger.error {err, project_id, user_id}, "error accessing web api" 
-				callback err
+				if response.statusCode == 429
```
das7pad:
```suggestion
			else if response.statusCode == 429
```

and remove the indent - might be easier to read

app/coffee/Errors.coffee
```diff
@@ -0,0 +1,9 @@
+RetryableError = (message) ->
```
das7pad: Maybe use your `overleaf/o-error` package?

gh2k: Given how critical real-time is, I'd rather not do this without more testing. cc @briangough

app/coffee/WebApiManager.coffee
gh2k: :+1: yeah, that's better

gh2k:
Amended the PR to add some tests and the suggested style improvement, and return an error code instead of a generic `retryable` datum as suggested in overleaf/web-internal#2115

app/coffee/WebApiManager.coffee
```diff
@@ -24,7 +25,10 @@ module.exports = WebApiManager =
 			return callback(error) if error?
 			if 200 <= response.statusCode < 300
 				callback null, data?.project, data?.privilegeLevel
+			else if response.statusCode == 429
+				logger.log(project_id, user_id, "rate-limit hit when joining project")
+				callback(new CodedError("rate-limit hit when joining project", "TooManyRequests"))
 			else
+				logger.error {err, project_id, user_id}, "error accessing web api"
```
das7pad: `err` is undefined here. It will be defined in the next line, but you might want to log the actual `error` instead?

gh2k: Good spot, thanks.

PR #81, opened 2019-10-04
Track the `isRestrictedUser` flag on clients

ShaneKilkelly:
### Description

Track the `isRestrictedUser` flag on clients  then, don't send new chat messages and new comments to those restricted clients.
We do this because we don't want to leak private information (email addresses
and names) to "restricted" users, those who have read-only access via a
shared token.

#### Related Issues / PRs

- Part of https://github.com/overleaf/issues/issues/2319
- Sibling of https://github.com/overleaf/web-internal/pull/2206

#### Potential Impact

It's real-time, so pretty high. We're now using Async in the broadcast-to-clients loop, which could have a performance impact.

#### Manual Testing Performed

- [x] Test that the editor works
- [x] Check that chat and comment messages are not sent to restricted clients
- [x] Check that chat and comment messages _are_ sent to other non-restricted clients

### Deployment

- Depends on https://github.com/overleaf/web-internal/pull/2206 to have any effect.

#### Deployment Checklist

- [ ] Keep an eye on metrics for real-time

#### Metrics and Monitoring

- Keep an eye on cpu and memory in real-time

### DIFF
--- a/app/coffee/WebApiManager.coffee
+++ b/app/coffee/WebApiManager.coffee
@@ -4,7 +4,7 @@ logger = require "logger-sharelatex"
 { CodedError } = require "./Errors"
 
 module.exports = WebApiManager =
-	joinProject: (project_id, user, callback = (error, project, privilegeLevel) ->) ->
+	joinProject: (project_id, user, callback = (error, project, privilegeLevel, isRestrictedUser) ->) ->
 		user_id = user._id
 		logger.log {project_id, user_id}, "sending join project request to web"
 		url = "#{settings.apis.web.url}/project/#{project_id}/join"
@@ -24,7 +24,7 @@ module.exports = WebApiManager =
 		}, (error, response, data) ->
 			return callback(error) if error?
 			if 200 <= response.statusCode < 300
-				callback null, data?.project, data?.privilegeLevel
+				callback null, data?.project, data?.privilegeLevel, data?.isRestrictedUser
 			else if response.statusCode == 429
 				logger.log(project_id, user_id, "rate-limit hit when joining project")
 				callback(new CodedError("rate-limit hit when joining project", "TooManyRequests"))
--- a/app/coffee/WebApiManager.coffee
+++ b/app/coffee/WebApiManager.coffee
@@ -24,7 +24,11 @@ module.exports = WebApiManager =
 		}, (error, response, data) ->
 			return callback(error) if error?
 			if 200 <= response.statusCode < 300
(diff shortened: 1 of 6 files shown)


ShaneKilkelly:
I've found a bunch more cases of messages we need to block. Will update...

das7pad:
What do you think about two rooms per project? One for public events (changes, join/leave...) and another for private events? restricted clients would join the public room only.

ShaneKilkelly:
That's a good idea! It'd be easier in the long run, rather than filtering messages on the one channel.

app/coffee/WebApiManager.coffee
```diff
@@ -24,7 +24,7 @@ module.exports = WebApiManager =
 		}, (error, response, data) ->
 			return callback(error) if error?
 			if 200 <= response.statusCode < 300
-				callback null, data?.project, data?.privilegeLevel
+				callback null, data?.project, data?.privilegeLevel, data?.isRestrictedUser
```
gh2k: Maybe out of scope, but I don't like this `data?` pattern here. If there's no data then surely this is an error, and it'll look like garbage if/when we decaf it.

app/coffee/WebsocketLoadBalancer.coffee
```diff
@@ -69,12 +82,27 @@ module.exports = WebsocketLoadBalancer =
...
+					room_id: message.room_id,
+					message_id: message._id,
+					socketIoClients: (client.id for client in clientList)
+				}, "distributing event to clients"
 				seen = {}
-				for client in clientList when not seen[client.id]
-					seen[client.id] = true
-					client.emit(message.message, message.payload...)
+				# Send the messages to clients async, don't wait for them all to finish
+				Async.eachSeries clientList
```
gh2k: Re the description, I can't see why this would be significantly slower.

Although... is there any mileage in making this `Async.each[Limit]` instead?

app/coffee/WebApiManager.coffee
ShaneKilkelly: good point, we should bail out early with an error if we don't get any data.

app/coffee/WebsocketLoadBalancer.coffee
ShaneKilkelly: I'm just a bit skittish about the extra function-call and callback overhead compared to the good old for-loop. We could certainly do an eachLimit to get a bit of concurrency.

PR #82, opened 2019-10-10
Add maxRetriesPerRequest setting for redis

gh2k:
### Description

As overleaf/document-updater#96 - add maxRetriesPerRequest setting for ioredis, to config.

#### Potential Impact

Significant, do not deploy


gh2k:
Is it real-time that we're setting up canary deployments for? In which case, that should probably go first.

henryoswald:
real time should be fine to deploy. We have done a dark deploy on the new nodepool already and it was fine.

Canary deployments are on doc updater.

PR #84, opened 2019-10-17
Bump to Node 10 and add error handlers for socket.io

gh2k:
### Description

Attempt to fix overleaf/issues#1833

This adds a global error handler which begins the drain+shutdown process, plus a `client.on "error"` handler, as suggested in the linked bug.

#### Related Issues / PRs

Fixes overleaf/issues#1833

### Review

I spent a long time trying to simulate errors so that I could write acceptance tests to check that the handlers are working. I did manage to test the global error handler, but not in a way that I could easily write a test for, as `chai` installs its own global exception handler, removing the one that we add.

I couldn't figure out how to get a client to simulate an error.

#### Potential Impact

All of Real-Time. We should be careful.

#### Manual Testing Performed

- [x] run chaosmonkey in a couple of tabs

### Deployment

We'd like to test out using a canary deployment for this.

#### Deployment Checklist

- [ ] Deploy with canary deployment
- [ ] Ramp up when things seem stable

#### Metrics and Monitoring

See the linked bug for the error message to look out for


das7pad:
We could run the app detached from the process running the tests.

On a quick glance I see just one blocker: there are two `sinon.spy` calls on app internals [1] - both for logger methods. The `logger.error` spy could be replaced with a fake sentry express server - which in turn would receive a payload upon calling `logger.error` (and waiting _a bit_). I assume the spy on `logger.warn` is an inverse to `logger.error` [2], as in checking that it logs on level warning instead of firing an error - this could use a similar detection with a fake sentry server.

---
[1] https://github.com/overleaf/real-time/blob/982530a8c52ac1793c4b71cf31c274fe1c4f7435/test/acceptance/coffee/LeaveDocTests.coffee#L18-L19
[2] https://github.com/overleaf/real-time/blob/982530a8c52ac1793c4b71cf31c274fe1c4f7435/test/acceptance/coffee/LeaveDocTests.coffee#L77-L84

henryoswald:
Let's get this out on Monday in prep for the patched node.js version on Tuesday. 

If the canary isn't behaving (it should be now) then let's just roll it out everywhere and keep an eye on it ready to roll back. The crashes before were not terrible, just one every hour or two hours, so after a quick roll back things won't be that bad for users.

app/coffee/Router.coffee
```diff
@@ -42,6 +42,12 @@ module.exports = Router =
 		app.post "/drain", httpAuth, HttpApiController.startDrain
 
 		session.on 'connection', (error, client, session) ->
+			client?.on "error", (err) ->
+				logger.err "socket.io client error", { err }
```
das7pad:
```suggestion
				logger.err { err }, "socket.io client error"
```

PR #85, opened 2019-10-22
Revert "Track the `isRestrictedUser` flag on clients"

gh2k:
Reverts overleaf/real-time#81

After deploying, I noticed that this behaviour doesn't seem quite right. I think this PR for real-time is still correct, but there are problems with the web UI.

Specifically, if the restricted user comes along to a project that already has comments, they see the comments as part of the UI. They can then reply to them and will see a spinner, whilst the comments are sent to the other users in the project.

Really, we need to add a PR to web to not show comments when the project is loaded, in order for this to be tied up properly. Otherwise, things just look a bit broken.

Example, from a read-write user:
image

From a read-only user:
image

### DIFF
--- a/app/coffee/WebApiManager.coffee
+++ b/app/coffee/WebApiManager.coffee
@@ -4,7 +4,7 @@ logger = require "logger-sharelatex"
 { CodedError } = require "./Errors"
 
 module.exports = WebApiManager =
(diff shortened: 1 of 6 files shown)


gh2k:
cc @ShaneKilkelly

PR #87, opened 2019-10-30
Filter "comments" if restricted user.

ShaneKilkelly:
### Description

Resurrection of https://github.com/overleaf/real-time/pull/81, and following on from its reversion in https://github.com/overleaf/real-time/pull/85.

Track the `isRestrictedUser` flag on clients  then, don't send new chat messages and new comments to those restricted clients.
We do this because we don't want to leak private information (email addresses
and names) to "restricted" users, those who have read-only access via a
shared token.

#### New changes since last PR

- Hide review comments from restricted users when they join a document

#### Related Issues / PRs

- Part of https://github.com/overleaf/issues/issues/2319
- Sibling of https://github.com/overleaf/web-internal/pull/2306 (current)
- Sibling of https://github.com/overleaf/web-internal/pull/2206 (old)

#### Potential Impact

It's real-time, so pretty high. We're now using Async in the broadcast-to-clients loop, which could have a performance impact.

#### Manual Testing Performed

- [x] Test that the editor works
- [x] Check that chat and comment messages are not sent to restricted clients
- [x] Check that chat and comment messages _are_ sent to other non-restricted clients
- [X] Check that comments are not received on the `joinDoc` event

### Deployment

- Depends on https://github.com/overleaf/web-internal/pull/2206 to have any effect.

#### Deployment Checklist

- [ ] Keep an eye on metrics for real-time

#### Metrics and Monitoring

- Keep an eye on cpu and memory in real-time

#### Who Needs to Know?
@jdleesmiller

### DIFF
--- a/app/coffee/WebApiManager.coffee
+++ b/app/coffee/WebApiManager.coffee
@@ -4,7 +4,7 @@ logger = require "logger-sharelatex"
 { CodedError } = require "./Errors"
 
 module.exports = WebApiManager =
-	joinProject: (project_id, user, callback = (error, project, privilegeLevel) ->) ->
+	joinProject: (project_id, user, callback = (error, project, privilegeLevel, isRestrictedUser) ->) ->
 		user_id = user._id
 		logger.log {project_id, user_id}, "sending join project request to web"
 		url = "#{settings.apis.web.url}/project/#{project_id}/join"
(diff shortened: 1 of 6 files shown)


ShaneKilkelly:
Hmm, seems I need to get github to include the old, reverted commits somehow.

das7pad:
You could revert the revert commit. A lazy approach would be the use of the `Revert` Button at the bottom of the PR that proposed the revert https://github.com/overleaf/real-time/pull/85#event-2732564264.

ShaneKilkelly:
Thanks for the tip @das7pad ! I've reverted the revert, so this is now ready for review.

PR #108, opened 2020-02-21
[misc] test/acceptance: add tests for the draining of connections

das7pad:
### Description

This PR adds tests for logic of the DrainManager.

There are other clients connected but there is no api endpoint to disconnect them at this time.
I do have patches available that implement this and will schedule them after this PR.
We can then lower the timeout in line 52 as the drain will schedule the reconnect for the two clients immediately -- currently the clients from other tests are asked to reconnect too, and there is a 1s delay in between the drain runs.

#### Related Issues / PRs

#### Potential Impact
Affects tests only.

### DIFF
--- a/test/acceptance/coffee/DrainManagerTests.coffee
+++ b/test/acceptance/coffee/DrainManagerTests.coffee
@@ -1,0 +1,72 @@
+RealTimeClient = require "./helpers/RealTimeClient"
+FixturesManager = require "./helpers/FixturesManager"
+
+expect = require("chai").expect
+
+async = require "async"
+request = require "request"
+
+Settings = require "settings-sharelatex"
+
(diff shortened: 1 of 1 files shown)


test/acceptance/coffee/DrainManagerTests.coffee
```diff
@@ -0,0 +1,72 @@
...
+	return null
+
+describe "DrainManagerTests", ->
+	before (done) ->
+		FixturesManager.setUpProject {
+			privilegeLevel: "owner"
+			project: {
+				name: "Test Project"
+			}
+		}, (e, {@project_id, @user_id}) => done()
```
gh2k: I hate this coffeescript syntax for dereferencing things, but it's where we are. ;-)

PR #101, opened 2020-02-17
Decaffeinate real-time

gh2k:
### Description

Decaffeinate real-time

#### Related Issues / PRs

Part of overleaf/web-internal#2555

### Review

The `decaffeinate.sh` script worked fine. There were a few extra ESLint/Prettier bits to do at the end, which always seems to happen with the script.

#### Manual Testing Performed

- [ ] Edit a document in multiple tabs

PR #113, opened 2020-02-24
[misc] synchronous client store using an Object at .ol_context

das7pad:
### Description
This is in preparation for socket.io v2 which does not have in memory client store anymore.

Upon connecting, clients will get a new Object at `.ol_context` for local attributes/context -- like the `user_id` and restricted status. It is deleted after the disconnect event is processed.

Fetching client store entries is much less verbose now using simple object unpacking.
`{project_id, user_id, is_restricted_user} = client.ol_context`

#### Related Issues / PRs

### Review
Some indents are twice as wide now to keep the git-diffs small. This will be corrected in the upcoming decaffeination process #101.

#### Potential Impact
High


app/coffee/WebsocketLoadBalancer.coffee
```diff
@@ -93,11 +100,9 @@ module.exports = WebsocketLoadBalancer =
 				Async.eachLimit clientList
```
briangough: @ShaneKilkelly while we are on the subject, what was the reason for introducing this `Async.eachLimit` here when sending the messages to the client? - I think the `client.emit` should always be non-blocking but something must have prompted adding this?

ShaneKilkelly: I think we needed to add this because we introduced a call to `Utils.getClientAttributes` inside the loop, on each client: https://github.com/overleaf/real-time/pull/81/files#diff-6888826980b5c4524173b5fa6eff8ecbR97

Now that we're removing the `getClientAttributes` call, I think we can go back to a plain for-loop.

briangough: Ah right, now I understand. That will be a nice cleanup to go back to a simple loop @das7pad.

das7pad: I filed a new PR which is targeting this PRs' branch: https://github.com/overleaf/real-time/pull/115

PR #117, opened 2020-02-27
[misc] upgrade socket.io to version 2.3.0

das7pad:
### Description
We are finally there: socket.io version **two** dot three dot zero.

##### getting clients

The function is async now and has a different signature/return value.
- v0: `io.socket.clients(ROOM) -> clientObjectListWithDuplicates`
- v2: `io.to(ROOM).clients((err, clientIdList) => {}) -> undefined`

As we fetch the client asynchronous, some of them may be gone by the time our callback is actually executed. Requiring us to validate each and every client again.
https://github.com/socketio/socket.io-adapter/blob/ae23c7ef4cd1b1793121b5af0376686a2ba2deea/index.js#L210

###### accessing clients

Clients are stored in an Object now. Available e.g. at `io.sockets.connected[SOCKET_ID]`

###### Rooms a client is in

The list at `CLIENT.rooms` includes the socket id at the first position.

###### Events

There is a new socket event `disconnecting` which is emitted just before leaving rooms. From the `disconnect` event we can not find out, which rooms the client was in anymore.

```
client.on "disconnect", () ->
    # This is called after leaving rooms.

client.on "disconnecting", () ->
    # This is called just before leaving rooms.
```

The socket is not tracked anymore by the time `disconnect` is fired. See this snipped from the DrainManagerTests -- which disconnect all clients -- with additional logging enabled. We no longer have to adjust the number of connected client for the metrics upon disconnecting.

<details>
Logging planted into both events respective:
```js
# into disconnect
console.error('disconnect', Object.keys(io.sockets.connected).length)
# into disconnecting
console.error('disconnecting', Object.keys(io.sockets.connected).length)
```

```
  DrainManagerTests
disconnecting 6
disconnect 5
disconnecting 5
disconnect 4
disconnecting 4
disconnect 3
disconnecting 3
disconnect 2
disconnecting 2
disconnect 1
disconnecting 1
disconnect 0
    ✓ should have disconnected all previous clients
```
</details>

---

##### middlewares

socket.io v2 supports middlewares, enabling much simpler session handling. The current flow is still functional and this is subject to a followup PR.

##### cookies

There is a default `io` cookie emitted now, which we do not need -- embedding the client id in the url continues to work as before.

##### client side JS serving

The client side JS is served minified by default now.

##### transports

There are only two transports left: `polling` and `websocket`.

---

#### Related Issues / PRs

### Review
Some more indent ticks were used to keep git diffs small -- this is again subject to change with the decaffeinate process.

See https://github.com/overleaf/web-internal/pull/2635 / use the `jpa-socket-io-v2` branch in web for testing.

#### Potential Impact
HIGH

#### Deployment Checklist

See https://github.com/overleaf/issues/issues/2757

#### Who Needs to Know?
@jdleesmiller


das7pad:
> First question: with socket.io v2 the client connects 3 times to the polling endpoint before connecting to the websocket. Is this expected?

I did not find any docs on this. We could list the websocket transport first -- and use polling as fallback. I was not able to change the sequence of the transports in the editor tho. I will experiment with some simpler apps and report back.

briangough:
It looks like the request behaviour is intentional:

> Socket.IO never assumes that WebSocket will just work, because in practice there’s a good chance that it won’t. Instead, it establishes a connection with XHR or JSONP right away, and then attempts to upgrade the connection to WebSocket. 

https://socket.io/blog/introducing-socket-io-1-0/#New-engine

das7pad:
From the same page:

> none of your users will have a degraded experience

_every user has a degraded startup experience_

das7pad:
I played around in browser stack with IE9 (the first browser I could find that does not support websockets) and the demo twitter stream from the socket.io webpage.

```
e=document.createElement('script');e.src="https://unpkg.com/socket.io-client@2.3.0/dist/socket.io.js";document.head.appendChild(e)
io.connect('https://socket-io-tweet-stream.now.sh/', {transports: ['websocket','polling'],forceNew:true})
```
Does not trigger any new connection. With no explicit transports, or a revered sequence, many polling requests are fired. So we definitely want to start with polling then (and the default transport selection).

---

In the socket.io v3 roadmap https://github.com/socketio/socket.io/issues/3250 there is an entry
 > Default to websocket, and use XHR polling as fallback

jdleesmiller:
Thanks for looking into it --- I think that behaviour made a lot more sense when socket.io 1 came out than it does now, but in any case I think we can probably live with it, at least initially.

briangough:
Another question, socket.io v2 says that it supports binary data, so maybe we can get rid of this encoding code:

https://github.com/overleaf/real-time/blob/cb50a91a4ac142363fd338c0fac672477359f1ab/app/coffee/WebsocketController.coffee#L102-L123

And the corresponding code in web: 

https://github.com/overleaf/web-internal/blob/a7bfce35c7b1e1b22fe347d9dd83502e0996f407/frontend/js/ide/editor/Document.js#L518-L540
https://github.com/overleaf/web-internal/blob/a7bfce35c7b1e1b22fe347d9dd83502e0996f407/frontend/js/ide/editor/ShareJsDoc.js#L34-L41

(I actually tried removing that in the current code with socket.io v 0.9  and it seems to work without  it -- so I'm not sure if it's even doing anything now!).

das7pad:
I can write up a PR with binary test data and see how this pans out.

TBH: I would rather keep the scope of the upgrade as small as possible. Let's do these changes separately -- who knows how the upgrade plays with all the different browsers out there.

app/coffee/DocumentUpdaterController.coffee
```diff
@@ -59,30 +59,27 @@ module.exports = DocumentUpdaterController =
...
+		io.to(doc_id).clients (err, clientIds) ->
+			if err?
+				return logger.err {room: doc_id, err}, "failed to get room clients"
+
+			for clientId in clientIds
+				continue unless io.sockets.connected.hasOwnProperty(clientId)
+				client = io.sockets.connected[clientId]
+				logger.warn err: error, doc_id: doc_id, client_id: client.id, "error from document updater, disconnecting client"
+				client.emit "otUpdateError", error, message
+				client.disconnect()
```
briangough: Using the v2 api I think we can write this whole function as

```
io.to(doc_id).emit("otUpdateError", error, message).disconnect()
```

Perhaps the same approach can be used elsewhere.

app/coffee/HttpController.coffee
```diff
@@ -11,22 +11,22 @@ module.exports = HttpController =
 			{project_id, user_id, first_name, last_name, email, connected_time} = ioClient.ol_context
 			client = {client_id, project_id, user_id, first_name, last_name, email, connected_time}
 			client.rooms = []
-			for name, joined of ioClient.manager.roomClients[client_id]
-				if joined and name != ""
-					client.rooms.push name.replace(/^\//, "") # Remove leading /
+			for name in Object.values(ioClient.rooms)
+				continue if name == ioClient.id
+				client.rooms.push name.replace(/^\//, "") # Remove leading /
```
briangough: I know it won't be here long but in coffeescript this can be written as 
```coffee
for name in Object.values(ioClient.rooms) when name isnt ioClient.id
```

app/coffee/DocumentUpdaterController.coffee
das7pad: `.disconnect()` can not be distributed to a room like a message can be broadcasted.

```
Uncaught TypeError: io.to(...).emit(...).disconnect is not a function
```

briangough: Ah yes, missed that in the logs -- it did disconnect all the clients, but not for the right reason!

briangough:
My other comment - not directly related to this code - is that based on  https://socket.io/docs/using-multiple-nodes/ we will need to keep sticky sessions for real-time for the polling fallback to work, so https://github.com/overleaf/google-ops/issues/645 will need to take account of that. cc @henryoswald

briangough:
> I can write up a PR with binary test data and see how this pans out.
> 
> TBH: I would rather keep the scope of the upgrade as small as possible. Let's do these changes separately -- who knows how the upgrade plays with all the different browsers out there.

Yes, better to keep this a minimal change.   In addition to handling binary data, it looks like we might also be able to make use of the socket.io-redis module for some cleanup in future as well.

app/coffee/DocumentUpdaterController.coffee
```diff
@@ -59,30 +59,27 @@ module.exports = DocumentUpdaterController =
...
+		source = update.meta?.source
+		if io.sockets.connected.hasOwnProperty(source)
+			emitter = io.sockets.connected[source]
+			logger.log {doc_id, version: update.v, source}, "distributing update to sender"
+			emitter.emit "otUpdateApplied", v: update.v, doc: update.doc
+		else
+			emitter = io
+		return if update.dup
+		logger.log {doc_id, version: update.v, source}, "distributing updates to clients"
+		emitter.to(doc_id).emit "otUpdateApplied", update
```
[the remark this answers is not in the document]
das7pad: > This code needs some comments to explain it.

I simplified the logic and added the comment in 818a3e509e5906a0d75b994b820cdef06fce68e2

---

> I think we could drop the hasOwnProperty as socket.io itself treats the id as safe to use directly as a property name? This will simplify the code a bit.

Done in be289b66821f1228a9bcdd6a29837c7cd8e62d64

app/coffee/HttpController.coffee
das7pad: Done in 09137f1fd68f7d26bfe78f90da941dbb7382927d

package.json
```diff
@@ -34,8 +34,8 @@
     "request": "^2.88.0",
     "session.socket.io": "^0.1.6",
     "settings-sharelatex": "^1.1.0",
-    "socket.io": "0.9.19",
-    "socket.io-client": "^0.9.16"
+    "socket.io": "2.3.0",
+    "socket.io-client": "2.3.0"
```
gh2k: Could `-client` be a devDependency?

test/unit/coffee/RoomManagerTests.coffee
```diff
@@ -182,16 +182,16 @@ describe 'RoomManager', ->
 					# second for the leave
 					@RoomManager._clientsInRoom
 						.withArgs(@client, @doc_id)
-						.onCall(0).returns(0)
-						.onCall(1).returns(0)
+						.onCall(0).yields(0)
+						.onCall(1).yields(0)
```
gh2k: Is it worth just saying `.yields` and dropping the `.onCall` calls, if we just want it to always yield 0?

app/coffee/DocumentUpdaterController.coffee
```diff
@@ -59,30 +59,24 @@ module.exports = DocumentUpdaterController =
...
-			if client.id == update.meta.source
-				logger.log doc_id: doc_id, version: update.v, source: update.meta?.source, "distributing update to sender"
-				client.emit "otUpdateApplied", v: update.v, doc: update.doc
-			else if !update.dup # Duplicate ops should just be sent back to sending client for acknowledgement
-				logger.log doc_id: doc_id, version: update.v, source: update.meta?.source, client_id: client.id, "distributing update to collaborator"
-				client.emit "otUpdateApplied", update
-		if Object.keys(seen).length < clientList.length
-			metrics.inc "socket-io.duplicate-clients", 0.1
-			logger.log doc_id: doc_id, socketIoClients: (client.id for client in clientList), "discarded duplicate clients"
+		source = update.meta?.source
```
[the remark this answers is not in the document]
das7pad: I am not sure whether it always exists. We are fine if `source` is undefined -> `sender=undefined` -> `sender=io`.

I was following the previous logging statement which had a check on `.meta`. Looking closer, other parts were accessing `.meta.source` straight away. So it _should_  be safe to clean this up.
https://github.com/overleaf/real-time/blob/3f6a461a2541091e06254084c4e660294510b0c0/app/coffee/DocumentUpdaterController.coffee#L72

package.json
das7pad: I am not sure. We want to serve the client side js in this particular version in production. Previously we diverged from `0.9.19` (backend)  and `0.9.16` (frontend).

socket.io depends on it in the matching version, but this might change at some-point?
https://github.com/socketio/socket.io/blob/47161a65d40c2587535de750ac4c7d448e5842ba/package.json#L31
The serving logic checks for our socket.io client blob first, before leveraging the `require` logic.
https://github.com/socketio/socket.io/blob/47161a65d40c2587535de750ac4c7d448e5842ba/lib/index.js#L105-L124

test/unit/coffee/RoomManagerTests.coffee
das7pad: As the comment explains above, these stubs are for two explicit calls: join, then leave.
We could remove the per argument spec too, but this explicit subbing is great for documentation.

package.json
gh2k: OK thanks - I thought it would have got it via webpack from the frontend.

app.coffee
```diff
@@ -23,13 +23,13 @@ CookieParser = require("cookie-parser")
...
-# work around frame handler bug in socket.io v0.9.16
-require("./socket.io.patch.js") 
 # Set up socket.io server
 app = express()
 
 server = require('http').createServer(app)
-io = require('socket.io').listen(server)
+io = require('socket.io')(server, {
+	cookie: false,
+})
```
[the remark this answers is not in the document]
das7pad: Done in 766dac398b972026ec0d8a07c1502bf7085fd9ce

app/coffee/DocumentUpdaterController.coffee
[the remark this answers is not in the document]
das7pad: Done in c3cb4d09e639b93d32b04b816a0526a655cfd955

gh2k:
I haven't run this, but from a code perspective (and having done some playing around with socket.io v2 in a hobby project over the last week or so) it looks good. :+1:
nicelydone

Will approve once I've had a look at the `web` end.

app/coffee/DocumentUpdaterController.coffee
```diff
@@ -74,9 +74,7 @@ module.exports = DocumentUpdaterController =
 			if err?
 				return logger.err {room: doc_id, err}, "failed to get room clients"
 
-			for clientId in clientIds
-				continue unless io.sockets.connected.hasOwnProperty(clientId)
-				client = io.sockets.connected[clientId]
+			for client in clientIds.map((id) -> io.sockets.connected[id]) when client # filter disconnected clients
```
briangough: What actually happens if we emit to or disconnect a client that has disconnected?  

Maybe this is just how it works but I feel like there should be some way to handle this without having to check clients explicitly.

das7pad: Here we are just receiving the list of client ids that were connected at the time of the previous process tick.

https://github.com/socketio/socket.io/blob/47161a65d40c2587535de750ac4c7d448e5842ba/lib/socket.js#L453
The disconnect logic removes the client from the `io.sockets.connected` Object immediately and we would try to call `undefined.emit(..)`.

briangough: Thanks for the link.   As I understand it `process.nextTick` will run after the current javascript finishes executing but _before_ the next event loop, so it is effectively a synchronous callback.  Does that make a difference? I think that would make the clients guaranteed to be valid and we could skip the check.

https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/#process-nexttick

das7pad: TIL, thanks. I will remove the checks then.

das7pad: [I've reworked the frontend code](https://github.com/overleaf/web-internal/compare/59d3ba0b4344734e323cfbd1bcb29618d5bd7b68..169d90cceb8f7e618c2ceb17da612d39519ef4c7) -- and was bitten by the early client cleanup :disappointed: 

Anyways here is the patch 8026003f8e9dab506d1a8a3e6e5f41188555a6bb.

app/coffee/DocumentUpdaterController.coffee
```diff
@@ -59,30 +59,24 @@ module.exports = DocumentUpdaterController =
...
-			logger.log doc_id: doc_id, socketIoClients: (client.id for client in clientList), "discarded duplicate clients"
+		source = update.meta.source
+		sender = io.sockets.connected[source]
+		if sender
+			logger.log {doc_id, version: update.v, source}, "distributing update to sender"
+			sender.emit "otUpdateApplied", v: update.v, doc: update.doc
+
+		return if update.dup
+		logger.log {doc_id, version: update.v, source}, "distributing updates to clients"
+		(sender || io).to(doc_id).emit "otUpdateApplied", update  # does not emit to sender if defined
```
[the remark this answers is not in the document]
das7pad: `sender` is the variable referencing the socket of the 'sender' of the update. `sender` may be undefined - see below when. But if it is defined, the `sender.to(doc_id).emit "....."` call does not emit the update back to the 'sender'.

This line covers three cases:

- the sender is connected to a different pod

  `sender` is undefined

- the sender was previously connected but by the time doc-updater confirms the update, the socket is gone already

   `sender` is undefined

- the sender is currently connected

   `sender` is defined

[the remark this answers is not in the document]
ShaneKilkelly: Good stuff. Could we add a comment to the code which clarifies these cases? I'm just a bit worried about coming back to this code in a few years and not being sure of what it's doing. Thanks!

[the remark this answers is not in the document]
das7pad: Is d4a34a641ff10a44821e8ba74ea57a41c2fec3ff sufficient?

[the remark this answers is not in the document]
ShaneKilkelly: Yes, thanks!

PR #118, opened 2020-03-05
[misc] mount socket.io on a configurable path

das7pad:
### Description
For testing of https://github.com/overleaf/google-ops/pull/859


gh2k:
Can we make this a setting? That way we can use an env var to put it in the right place without having to modify the code later.

das7pad:
> Can we make this a setting? That way we can use an env var to put it in the right place without having to modify the code later.

Done. _But it will make testing locally more complex._

PR #119, opened 2020-03-20
[misc] let proxys observe an upcoming shutdown before starting to drain

das7pad:
### Description

As per the title. Proxys in front of `real-time` may poll the health status and push traffic to the pod until it shows a failing state. Reconnect requests should not go to the same pod. We have to wait for the proxy to observe that a given pod is going down soon, before we can start with draining.

Default is do not delay the draining. Adjust via the environment variable `STATUS_CHECK_INTERVAL`.

#### Related Issues / PRs
https://github.com/overleaf/dev-environment/pull/324

#### Potential Impact
Low

#### Manual Testing Performed

- see https://github.com/overleaf/dev-environment/pull/324

### DIFF
--- a/app.coffee
+++ b/app.coffee
@@ -125,9 +125,14 @@ drainAndShutdown = (signal) ->
 		return
 	else
 		Settings.shutDownInProgress = true
-		logger.warn signal: signal, "received interrupt, starting drain over #{shutdownDrainTimeWindow} mins"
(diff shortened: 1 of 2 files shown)


das7pad:
> I assume this is unavoidable and the main goal is to avoid the large number of reconnections coming in from the drain - is that correct?

Router: Yes, there is no point in serving a new user for a few seconds.

I only want to avoid the reconnects to the same pod, due to persistence. New connections may bounce immediately, but that's OK.

PR #120, opened 2020-03-24
[misc] limit update size

das7pad:
### Description

We should limit inbound socket.io traffic.

Large updates get pushed to redis, pulled by doc-updater and discarded then.
The frontend will not receive a confirmation for the update and will push the
 update again and again, until it gets disconnected by real-time when the
 rejection from doc-updater comes in as an 'otUpdateError'.

Related and also included in this PR:

We can push an update to doc-updater that is larger than a single doc is
 allowed to be (2mb). The confirmation from doc-updater would get rejected
however, the frontend will never see the confirmation and send the update
 again and again.

#### Review

The limit is set to two doc-sizes (delete and insert new full doc) plus the
 maximum size of ranges: 2*2mb + 3mb.

In theory we should add JSON encoding overhead to the limit.
But I could not find a way to max out the current limit to its full potential,
 meaning full 2mb doc replacement and generating 3mb worth of ranges,
 using our frontend -- and not a rough third-party client.

_Rejecting a 1.5mb range is blocked at the frontend with a doc-is-too-large
 and out-of-sync modal, real-time does not see the update -- bug?_

#### Potential Impact

High. 

#### Manual Testing Performed

I added unit/acceptance tests.

##### allowed

- create a payload of just below 2mb
- paste it into a project/document
- copy everything in the doc
- paste the payload over it (replace it)
- the update should be allowed

##### rejected

- repeat the previous steps in a loop
- eventually more than two payloads are in a single update

  (delete, insert, delete, insert)

- get an out-of-sync message

#### Metrics and Monitoring

new metric: `update_too_large`
https://github.com/overleaf/real-time/blob/7572a8a9504a13571a15fc466ff0f96924f351b1/app/coffee/WebsocketController.coffee#L207-L209

### DIFF
--- a/app/coffee/SafeJsonParse.coffee
+++ b/app/coffee/SafeJsonParse.coffee
@@ -3,8 +3,8 @@ logger = require "logger-sharelatex"
 
 module.exports =
 	parse: (data, callback = (error, parsed) ->) ->
-		if data.length > (Settings.max_doc_length or 2 * 1024 * 1024)
-			logger.error {head: data.slice(0,1024)}, "data too large to parse"
+		if data.length > Settings.maxUpdateSize
+			logger.error {head: data.slice(0,1024), length: data.length}, "data too large to parse"
 			return callback new Error("data too large to parse")
 		try
 			parsed = JSON.parse(data)
--- a/app/coffee/WebsocketController.coffee
+++ b/app/coffee/WebsocketController.coffee
@@ -1,5 +1,6 @@
 logger = require "logger-sharelatex"
 metrics = require "metrics-sharelatex"
+settings = require "settings-sharelatex"
 WebApiManager = require "./WebApiManager"
 AuthorizationManager = require "./AuthorizationManager"
 DocumentUpdaterManager = require "./DocumentUpdaterManager"
@@ -202,6 +203,23 @@ module.exports = WebsocketController =
 		Utils.getClientAttributes client, ["user_id", "project_id"], (error, {user_id, project_id}) ->
 			return callback(error) if error?
 			return callback(new Error("no project_id found on client")) if !project_id?
+
+			updateSize = JSON.stringify(update).length
+			if updateSize > settings.maxUpdateSize
+				metrics.inc "update_too_large"
+				logger.warn({user_id, project_id, doc_id, updateSize}, "update is too large")
+
+				# mark the update as received -- the client should not send it again!
+				callback()
+
+				# trigger an out-of-sync error
+				message = {project_id, doc_id, error: "update is too large"}
+				setTimeout () ->
+					client.emit "otUpdateError", message.error, message
+					client.disconnect()
+				, 100
+				return
+
 			WebsocketController._assertClientCanApplyUpdate client, doc_id, update, (error) ->
 				if error?
 					logger.warn {err: error, doc_id, client_id: client.id, version: update.v}, "client is not authorized to make update"
--- a/app/coffee/WebsocketController.coffee
+++ b/app/coffee/WebsocketController.coffee
@@ -204,22 +204,6 @@ module.exports = WebsocketController =
 			return callback(error) if error?
 			return callback(new Error("no project_id found on client")) if !project_id?
 
-			updateSize = JSON.stringify(update).length
-			if updateSize > settings.maxUpdateSize
-				metrics.inc "update_too_large"
-				logger.warn({user_id, project_id, doc_id, updateSize}, "update is too large")
-
-				# mark the update as received -- the client should not send it again!
-				callback()
-
-				# trigger an out-of-sync error
-				message = {project_id, doc_id, error: "update is too large"}
-				setTimeout () ->
-					client.emit "otUpdateError", message.error, message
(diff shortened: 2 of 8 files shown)


app/coffee/WebsocketController.coffee
```diff
@@ -202,6 +203,23 @@ module.exports = WebsocketController =
 		Utils.getClientAttributes client, ["user_id", "project_id"], (error, {user_id, project_id}) ->
 			return callback(error) if error?
 			return callback(new Error("no project_id found on client")) if !project_id?
+
+			updateSize = JSON.stringify(update).length
```
henryoswald: Although JSON.stringify is very fast this is a very hot code path, when we have done even a bit more processing here before we do see a decent uptick in cpu usage, what if we pushed the check into `DocumentUpdaterManager.queueChange` where we already stringify the update and do some sanity checks on the data looking for null bytes.

das7pad: I did that in a previous iteration, but thought we want to reject a super large payload asap.
E.g. `update.op` could be very large array which produces a very busy loop here:
https://github.com/overleaf/real-time/blob/c3d76760a6f347b81bdcc8b1426265e074017d8d/app/coffee/WebsocketController.coffee#L238-L241

das7pad: > do some sanity checks on the data looking for null bytes.

I fell for that too: This is a memory corruption check, not an actual check for null bytes in the update.
-> https://github.com/overleaf/document-updater/pull/110#issuecomment-580297820

gh2k: I would tend to trust Henry's judgement on performance issues here. He's usually been right in my experience... If it's not harmful to reject the update later I would do that.

I think it's better for the loop to be slower just in the case of huge documents, than to stringify on every request, even small ones.

das7pad: Alright, I will move the check.

gh2k:
Caveat: Realtime isn't really my area of expertise.

Just to be sure I understand correctly: DocUpdater does remove the update when it is received, but the key size can because the client re-sends the update more frequently than DocUpdater can reject it?

app/coffee/WebsocketController.coffee
das7pad: Done in 8127b75

das7pad:
> Just to be sure I understand correctly: DocUpdater does remove the update when it is received, but the key size can [build up] because the client re-sends the update more frequently than DocUpdater can reject it?

Yes.

<details>

Doc-updater will process 8 updates at a time: 
https://github.com/overleaf/document-updater/blob/7d59b0d4ac03c2187a554a8596f29dcfc6e55284/app/coffee/RealTimeRedisManager.coffee#L13
After these 8 blobs, it will continue with the next set of 8 -- when no error was emitted, this can be a problem, see below.

So there can be some delay, causing timeouts and resending from the frontend for large/noisy projects. Which will in turn cause even larger queues.


</details>

---

I reviewed the logic in doc-updater and there is a possibility that updates do not get processed in case of a crash of the pod -- or when some update processing errors out.  It goes like this:

<details>

- real-time queues updates continuously
- doc-updater 1: sees the queue entry and will pull 8 updates
- doc-updater 2: also sees the project/docs with pending ops, but will not get the lock for processing, which doc-updater 1 holds. So it moves on
- doc-updater 1: some update processing errors out/crashes
- user disconnects (front end will not send unconfirmed op again)
- real-time stops sending updates (no new entries appear in the queue)
- updates remain in redis, until ...
- real-time sends new updates
- some doc-updater pulls old updates first, moves to new ones there after
</details>

app/coffee/SafeJsonParse.coffee
```diff
@@ -3,8 +3,8 @@ logger = require "logger-sharelatex"
 
 module.exports =
 	parse: (data, callback = (error, parsed) ->) ->
-		if data.length > (Settings.max_doc_length or 2 * 1024 * 1024)
-			logger.error {head: data.slice(0,1024)}, "data too large to parse"
+		if data.length > Settings.maxUpdateSize
+			logger.error {head: data.slice(0,1024), length: data.length}, "data too large to parse"
```
gh2k: 1k seems a lot to log... I doubt we need that much?

config/settings.defaults.coffee
```diff
@@ -52,6 +52,11 @@ settings =
 	
 	max_doc_length: 2 * 1024 * 1024 # 2mb
 
+	# combine
+	# max_doc_length (2mb see above) * 2 (delete + insert)
+	# max_ranges_size (3mb see MAX_RANGES_SIZE in doc-updater) 
+	maxUpdateSize: parseInt(process.env['MAX_UPDATE_SIZE']) or 7 * 1024 * 1024
```
henryoswald: is 7Mb from 3mb * 2 + some headroom?

das7pad: It's 3mb (ranges) + 2 * 2mb (doc lines). I could not max out both doc lines and ranges. Anyways I will add some headroom.

PR #126, opened 2020-04-21
[Router] gracefully set and do not reset the ol_context

das7pad:
### Description

For https://digital-science.slack.com/archives/G6Q45PYQL/p1587468755015700

There seem to be events coming in after the socket entered the 'disconnecting' state (before 'disconnect'). It should be OK to leave the `ol_context` on the client and let v8 clean it up as it destructs the client.

#### Potential Impact

Low, race condition see slack.

### DIFF
--- a/app/coffee/Router.coffee
+++ b/app/coffee/Router.coffee
@@ -41,7 +41,9 @@ module.exports = Router =
 		app.post "/client/:client_id/disconnect", httpAuth, HttpApiController.disconnectClient
 
 		session.on 'connection', (error, client, session) ->
-			client.ol_context = {} unless client.ol_context
+			# init client context, we may access it in Router._handleError before
+			#  setting any values
+			client.ol_context = {}
 
 			client?.on "error", (err) ->
 				logger.err { clientErr: err }, "socket.io client error"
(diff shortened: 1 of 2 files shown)


app/coffee/Router.coffee
```diff
@@ -41,7 +41,7 @@ module.exports = Router =
 		app.post "/client/:client_id/disconnect", httpAuth, HttpApiController.disconnectClient
 
 		session.on 'connection', (error, client, session) ->
-			client.ol_context = {}
+			client.ol_context = {} unless client.ol_context
```
briangough, line 44: `client.ol_context ?= {}` works in coffeescript (it's the idiomatic way to do this).

maybe add a comment here that the client can be reconnecting, hence we can preserve the context in that case.

I think it would be good to always reset the `ol_context` with `client.ol_context = {}` at [WebsocketController.coffee#L28](https://github.com/overleaf/real-time/blob/8127b75f4fe92d72c7690e583815dd57bf5591bf/app/coffee/WebsocketController.coffee#L28) just before the context is populated.  That  would  ensure the context is always fully reinitialised on joining a project and no properties are carried over in future (I think this is the only place we set things at the moment so it's a defensive change).

das7pad: https://github.com/overleaf/real-time/pull/126/commits/bb33a8970a4960af734f7e5e1ee93d1b3ac136c7 I have reverted the `client.ol_context ?= {}` line. Running it multiple times would result in multiple event handlers too. This is not the case from my testing.

https://github.com/overleaf/real-time/pull/126/commits/7427122b0328002a4c61fe1b9b6ee6afb5c4aeeb The context is now reset in `joinProject` as suggested.

PR #107, opened 2020-02-21
[misc] test/acceptance: ReceiveUpdateTests: add 2nd project/3rd client

das7pad:
### Description
This PR extends the `ReceiveUpdateTests` in adding a 2nd project and a 3rd user that joins this 2nd project but not the first one.

The intend is to check for cross project leaks from the message fan-out.

Sending messages and errors from doc-updater is tested -- the errors are paired with a disconnect.

#### Related Issues / PRs
https://github.com/overleaf/issues/issues/2757

#### Potential Impact
Affects tests only.

PR #128, opened 2020-04-28
[misc] hack around process next tick

das7pad:
### Description

Implement https://digital-science.slack.com/archives/G6256GXGS/p1587995985009900?thread_ts=1587995558.008500&cid=G6256GXGS

Getting a client list and processing the contents of it is now safe -- the clients in the list are all connected and there are no other clients that may joined/left the room as we process the list.

This fixes the race condition in socket.io-adapter of calculating the room list and returning the list via `process.nextTick`.
https://github.com/socketio/socket.io-adapter/blob/ae23c7ef4cd1b1793121b5af0376686a2ba2deea/index.js#L210

Given that we check the connected state of all clients in the new helper `RoomManager.getClientsInRoomSync` it is safe to revert #127 .

#### Related Issues / PRs

#127 

### Review

The unit tests are probably useless when upgrading socket.io, but I added them anyways.

#### Potential Impact

High. `getClientsInRoomSync` is deciding who can see which message.

#### Manual Testing Performed

- acceptance tests were extended as part of the socket.io v2 upgrade -- including cross project message leak checks: https://github.com/overleaf/real-time/pull/107

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,2 @@
+[app/coffee/RoomManager.coffee]
+indent_style = space
--- a/app/coffee/DocumentUpdaterController.coffee
+++ b/app/coffee/DocumentUpdaterController.coffee
@@ -80,7 +80,7 @@ module.exports = DocumentUpdaterController =
 			if err?
 				return logger.err {room: doc_id, err}, "failed to get room clients"
 
-			for client in clientIds.map((id) -> io.sockets.connected[id]) when client # filter disconnected clients
+			for client in clientIds.map((id) -> io.sockets.connected[id])
 				logger.warn err: error, doc_id: doc_id, client_id: client.id, "error from document updater, disconnecting client"
 				client.emit "otUpdateError", error, message
 				client.disconnect()
--- a/app/coffee/DocumentUpdaterController.coffee
+++ b/app/coffee/DocumentUpdaterController.coffee
@@ -76,7 +76,7 @@ module.exports = DocumentUpdaterController =
 		(sender || io).to(doc_id).emit "otUpdateApplied", update
 
 	_processErrorFromDocumentUpdater: (io, doc_id, error, message) ->
-		io.to(doc_id).clients (err, clientIds) ->
+		RoomManager.getClientsInRoomPseudoAsync io, doc_id, (err, clientIds) ->
 			if err?
 				return logger.err {room: doc_id, err}, "failed to get room clients"
 
--- a/app/coffee/RoomManager.coffee
+++ b/app/coffee/RoomManager.coffee
@@ -87,8 +87,7 @@ module.exports = RoomManager =
             metrics.gauge "room-listeners", RoomEvents.eventNames().length
 
     _clientsInRoom: (client, room, cb) ->
-        client.server.in(room).clients (err, clients) ->
-            cb(clients?.length || 0)
+        cb(@getClientsInRoomSync(client.server, room).length)
 
     _roomsClientIsIn: (client) ->
         # skip the socket id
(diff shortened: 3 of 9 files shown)


jdleesmiller:
Without knowing too much about this... is this a change that we should look to push upstream?

das7pad:
> Without knowing too much about this... is this a change that we should look to push upstream to socket.io?

I am not sure about this one. The idea to implement this in an asynchronous fashion was presumably to let other operation get some CPU time after we used a lot from looping over all connected users in a room -- and maybe to provide a similar interface as the other adapters (e.g. with redis you would get the list of clients after a few RTT to the redis backend).

There are a few other places that use `process.nextTick` too, so this is a broader design fault I'd say. E.g. the bubbling up from underlaying socket read -> message decoding -> ... -> our event handler has multiple `process.nextTick` in between. This could lead to other races like getting a `joinProject` request after a `disconnect` event. `joinProject` is a regular message which requires decoding (at least 2 `process.nextTick`) and the disconnect is fired immediately (still at least 1 `process.nextTick`). 

<details>
<summary> joinProject bubbling </summary>

https://digital-science.slack.com/archives/G6Q45PYQL/p1587639247027100?thread_ts=1587634527.021500&cid=G6Q45PYQL

<pre>
All messages/events (packages from the client) seem to be processed via TWO `process.nextTick` callbacks:
- #1 https://github.com/socketio/socket.io/blob/ac945d1eba2c0e126c6750d5eccbdb861e0abc56/lib/client.js#L213-L215
- https://github.com/socketio/socket.io/blob/1decae341c80c0417b32d3124ca30c005240b48a/lib/socket.js#L328
- https://github.com/socketio/socket.io/blob/1decae341c80c0417b32d3124ca30c005240b48a/lib/socket.js#L364
- #2 https://github.com/socketio/socket.io/blob/1decae341c80c0417b32d3124ca30c005240b48a/lib/socket.js#L523-L531
</pre>
</details>

<details>
<summary> disconnect bubbling </summary>

https://digital-science.slack.com/archives/G6Q45PYQL/p1587639459027400?thread_ts=1587634527.021500&cid=G6Q45PYQL
<pre>
Client disconnect from a single `process.nextTick` callback:
- #1 https://github.com/socketio/socket.io/blob/ac945d1eba2c0e126c6750d5eccbdb861e0abc56/lib/client.js#L213-L215
- https://github.com/socketio/socket.io/blob/1decae341c80c0417b32d3124ca30c005240b48a/lib/socket.js#L340
- https://github.com/socketio/socket.io/blob/1decae341c80c0417b32d3124ca30c005240b48a/lib/socket.js#L418
- https://github.com/socketio/socket.io/blob/1decae341c80c0417b32d3124ca30c005240b48a/lib/socket.js#L453
</pre>
</details>

---

While I am writing this, I noticed that this (very likely) introduces a redis pub/sub subscribe leak -- as in we listen for messages that no connected client will receive. We are not cancelling the pub/sub subscribe for a `joinProject` that came in after a disconnect procedure ran.

@briangough  suggested to decouple the unsubscribe from the users disconnect procedure  -- e.g. in running a detached loop to look for empty rooms.

This is actually required now and I will look into it tomorrow. I have already started looking into cancelling the various request when we detect that a user disconnected as we process their event -- e.g. in `joinProject` we can skip fetching a bunch of data from web, when the user does not receive the result anyway.

das7pad:
This makes me think once again -- is socket.io v2 actually production ready? Probably not.

jdleesmiller:
OK, thanks for the explanation. I can see why those would need to be async in general, as you say.

Re redis pub/sub leak: would we be able to confirm that by looking at metrics from redis (or somewhere else)? Happy to discuss production readiness on a call if that would be helpful.

das7pad:
> Re redis pub/sub leak: would we be able to confirm that by looking at metrics from redis (or somewhere else)?

I do not think so -- these are quite rare  and there was one case in two days with beta-user canary traffic. We are also not tracking room sizes or updates on a project level, so there is no way to tell whether a room has active clients from metrics (which we could then correlate with metrics from redis).

das7pad:
I did some digging in git blame and the related PRs (and linked issues) do not confirm my assumption on why `process.nextTick` was used. These PRs are from 2015 and 2016  -- it is very likely that NodeJS (v4) processed a single network message in an event loop cycle at that time. 

`joinProject` chain:
1. A patch for user event handlers that throw `https://github.com/socketio/socket.io/pull/2682`
2. Initial support for middlewares (no comment on the delayed invocation of the user handler): `https://github.com/socketio/socket.io/pull/2306`

 `disconnect`
1. see 1. above

The question now is: propose the removal of `process.nextTick` upstream (-> socket.io v3) or move on with v2 and continue hacking around the race conditions.
cc @jdleesmiller @briangough

briangough:
I think we should move ahead with v2 and handle any edge-cases for out of order messages (e.g. don't assume that we are connected when we get joinProject, and have some background cleanup of empty rooms).

jdleesmiller:
> move ahead with v2

👍 That seems like the way to go. I think you could raise this as an issue upstream to see what the response is, if it is a regression since node 4+.

das7pad:
Alright I will create followup PRs then.

PR #129, opened 2020-04-30
[misc] get the list of rooms a client is in in a synchronous fashion

das7pad:
### Description

Followup for #128 

`client.rooms` is not safe to use.

<details>
the implementation in socket.io is prone to race conditions:
The adapter will get updated immediately from joining, but the client
 local state (.rooms) is updated via process.nextTick.
Given that NodeJS can process multiple network events in the same event
 loop cycle, we can join a project and disconnect in the same cycle.
We will not pick up the room (we just joined) for leaving.
</details>

#### Related Issues / PRs

#128 

#### Potential Impact

Low.

### DIFF
--- a/app/coffee/RoomManager.coffee
+++ b/app/coffee/RoomManager.coffee
@@ -122,7 +122,5 @@ module.exports = RoomManager =
         #  cycle. We would not pick up the room (we just joined) for leaving.
         roomRegistry = client.server.sockets.adapter.sids
         return [] unless roomRegistry.hasOwnProperty(client.id)
-        rooms = Object.keys(roomRegistry[client.id])
-        # skip the socket id
-        rooms.splice(rooms.indexOf(client.id), 1)
+        rooms = (room for room in Object.keys(roomRegistry[client.id]) when room isnt client.id)  # exclude the client socket entry
         return rooms
--- a/app/coffee/RoomManager.coffee
+++ b/app/coffee/RoomManager.coffee
@@ -90,11 +90,10 @@ module.exports = RoomManager =
         cb(@getClientsInRoomSync(client.server, room).length)
 
     _roomsClientIsIn: (client) ->
-        # skip the socket id
-        return Object.keys(client.rooms).slice(1)
+        return @getRoomsClientIsInSync(client)
 
     _clientAlreadyInRoom: (client, room) ->
-        return client.rooms.hasOwnProperty(room)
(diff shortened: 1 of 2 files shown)


app/coffee/RoomManager.coffee
```diff
@@ -113,3 +112,9 @@ module.exports = RoomManager =
     # calling it asynchronously would lead to race conditions -- see above
     getClientsInRoomPseudoAsync: (io, room, cb) ->
         cb(null, RoomManager.getClientsInRoomSync(io, room))
+
+    getRoomsClientIsInSync: (client) ->
+        roomRegistry = client.server.sockets.adapter.sids
+        return [] unless roomRegistry.hasOwnProperty(client.id)
+        # skip the socket id
+        return Object.keys(roomRegistry[client.id]).slice(1)
```
briangough, line 120: I see it was like this already but is there a way to exclude the client's own socket robustly, rather than depending on its position in the array returned by `Object.keys`?  I am not sure whether the order is guaranteed -- looking at MDN (which may not be 100% reliable) says `Object.keys` returns the same order as `for...in`, and that `for...in` order is implementation-dependent.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

das7pad: The id is on first position in Node 12 and 14 too -- but I can make it more explicit.

app/coffee/RoomManager.coffee
```diff
@@ -113,3 +112,17 @@ module.exports = RoomManager =
...
+        # The adapter will get updated immediately from joining, but the
+        #  client local state (.rooms) is updated via process.nextTick.
+        # Given that NodeJS can process multiple network events in the same
+        #  event loop cycle, we can join a project and disconnect in the same
+        #  cycle. We would not pick up the room (we just joined) for leaving.
+        roomRegistry = client.server.sockets.adapter.sids
+        return [] unless roomRegistry.hasOwnProperty(client.id)
+        rooms = Object.keys(roomRegistry[client.id])
+        # skip the socket id
+        rooms.splice(rooms.indexOf(client.id), 1)
```
briangough, line 127: While we are still in coffescript let's make it idiomatic
```suggestion
        rooms = (room for room in Object.keys(roomRegistry[client.id]) when room isnt client.id)  # exclude the client socket entry
```

PR #130, opened 2020-05-01
[misc] clear on subscribe failure

das7pad:
### Description

For https://github.com/overleaf/issues/issues/3002

Upon detecting a redis pub/sub subscribe failure we should discard the 'subscribed' state for the entity and let the next request start from scratch.

Currently based onto #129 

#### Related Issues / PRs

### Review

`sinon` < 2 does not support promises. So I went the long way of upgrading it for proper stub support https://github.com/overleaf/real-time/commit/3c92b937f9430175d7c51660e03c507635448e88. This required some fixes to the way we check error messages -- in fact sinon did not check the error message and lots of tests failed after the upgrade https://github.com/overleaf/real-time/commit/d9570fee70701a5f431c39fdbec5f8bc5a7843fe.

I could not get the unhandled promise rejection warning in the unit tests to disappear, even tho I added a `catch` handler. Presumably with the decaff process we can rewrite this into an `async` function (which is technically supported by coffee-script, but not in our ancient version). I left a TODO line for that.

#### Potential Impact

Low. Subscribing to a redis channel twice is considered safe.

#### Metrics and Monitoring

New metric `subscribe.failed.#{baseChannel}` (base channel is `applied-ops` or `editor-events`)

### DIFF
--- a/app/coffee/ChannelManager.coffee
+++ b/app/coffee/ChannelManager.coffee
@@ -31,13 +31,11 @@ module.exports = ChannelManager =
             clientChannelMap.set(channel, subscribePromise)
             logger.log {channel}, "subscribed to new channel"
             metrics.inc "subscribe.#{baseChannel}"
-            return new Promise (resolve, reject) ->
-                subscribePromise.then(resolve)
-                subscribePromise.catch (err) ->
-                    metrics.inc "subscribe.failed.#{baseChannel}"
-                    # clear state
-                    clientChannelMap.delete(channel)
-                    reject(err)
+            subscribePromise.catch () ->
+                metrics.inc "subscribe.failed.#{baseChannel}"
+                # clear state
+                clientChannelMap.delete(channel)
+            return subscribePromise
 
     unsubscribe: (rclient, baseChannel, id) ->
         clientChannelMap = @getClientMapEntry(rclient)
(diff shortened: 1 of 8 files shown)


app/coffee/ChannelManager.coffee
```diff
@@ -31,7 +31,13 @@ module.exports = ChannelManager =
...
             logger.log {channel}, "subscribed to new channel"
             metrics.inc "subscribe.#{baseChannel}"
-            return subscribePromise
+            return new Promise (resolve, reject) ->
+                subscribePromise.then(resolve)
+                subscribePromise.catch (err) ->
+                    metrics.inc "subscribe.failed.#{baseChannel}"
+                    # clear state
+                    clientChannelMap.delete(channel)
+                    reject(err)
```
briangough, line 40: could we do this without creating a new promise?

```
subscribePromise.catch (err) -> 
    # handle failure case
    metrics.inc ..
    clientChannelMap.delete(channel)
return subscribePromise
```

das7pad: Good idea, done in https://github.com/overleaf/real-time/pull/130/commits/e8d18b271a73ecf6559a592046eeebbf8e8c7017

PR #132, opened 2020-05-12
[ChannelManager] cleanup from redis subscribe error more carefully

das7pad:
### Description

There is a minor race condition with the redis pub/sub subscribe cleanup on error:

Take this sequence: subscribe -> unsubscribe -> subscribe

The unsubscribe clears the subscribe map and the second subscribe will
 insert its promise into the map. We should not cleanup the subscribe
 promise when the first subscribe errors.

Without this patch, we are doing another subscribe request for a 3rd subscribe call.

#### Related Issues / PRs

### Review

I added another commit to suppress the unhandledRejection warnings -- we do handle them, but from a sync function and moving to async functions will magically get rid of them when we are there.

#### Potential Impact

Low. Code path is reachable from failing subscribes, which should not happen.

#### Manual Testing Performed

- added unit test

### DIFF
--- a/app/coffee/ChannelManager.coffee
+++ b/app/coffee/ChannelManager.coffee
@@ -33,8 +33,10 @@ module.exports = ChannelManager =
             metrics.inc "subscribe.#{baseChannel}"
             subscribePromise.catch () ->
                 metrics.inc "subscribe.failed.#{baseChannel}"
-                # clear state
-                clientChannelMap.delete(channel)
+                # clear state on error, following subscribes should retry
+                # new subscribe requests can overtake, skip cleanup then
+                if clientChannelMap.get(channel) is subscribePromise
+                    clientChannelMap.delete(channel)
             return subscribePromise
 
     unsubscribe: (rclient, baseChannel, id) ->
--- a/test/unit/coffee/ChannelManagerTests.coffee
+++ b/test/unit/coffee/ChannelManagerTests.coffee
@@ -34,7 +34,6 @@ describe 'ChannelManager', ->
 				@rclient.subscribe.called.should.equal false
 
 		describe "when subscribe errors", ->
-			# TODO(das7pad): rework after decaff -- our coffee-script version does not like async/await
 			beforeEach (done) ->
 				@rclient.subscribe = () ->
 					return new Promise (resolve, reject) ->
@@ -43,7 +42,7 @@ describe 'ChannelManager', ->
 				@rclient.subscribe = sinon.stub().resolves()
 				p.then () ->
 					done(new Error('should not subscribe but fail'))
-				p.catch (err) =>
+				.catch (err) =>
 					err.message.should.equal "some redis error"
 					@ChannelManager.getClientMapEntry(@rclient).has("applied-ops:1234567890abcdef").should.equal false
 					@ChannelManager.subscribe @rclient, "applied-ops", "1234567890abcdef"
--- a/test/unit/coffee/ChannelManagerTests.coffee
+++ b/test/unit/coffee/ChannelManagerTests.coffee
@@ -53,6 +53,29 @@ describe 'ChannelManager', ->
 				@rclient.subscribe.called.should.equal true
 				@ChannelManager.getClientMapEntry(@rclient).has("applied-ops:1234567890abcdef").should.equal true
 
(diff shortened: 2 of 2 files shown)


test/unit/coffee/ChannelManagerTests.coffee
```diff
@@ -33,26 +33,53 @@ describe 'ChannelManager', ->
...
 			# TODO(das7pad): rework after decaff -- our coffee-script version does not like async/await
-			beforeEach (done) ->
-				@rclient.subscribe = () ->
-					return new Promise (resolve, reject) ->
-						setTimeout((() -> reject(new Error("some redis error"))), 1)
-				p = @ChannelManager.subscribe @rclient, "applied-ops", "1234567890abcdef"
-				@rclient.subscribe = sinon.stub().resolves()
-				p.then () ->
-					done(new Error('should not subscribe but fail'))
-				p.catch (err) =>
```
briangough, line 46: I think the unhandled rejection is coming from the p.then(), which returns a promise which is not handled.

If you change it to 

```
p.then () ->
	done(new Error('should not subscribe but fail'))
.catch (err) =>
	err.message.should.equal "some redis error"
        ....
```
Then the warning will go away and you don't need the changes in the first commit.

app/coffee/ChannelManager.coffee
```diff
@@ -33,8 +33,10 @@ module.exports = ChannelManager =
             metrics.inc "subscribe.#{baseChannel}"
             subscribePromise.catch () ->
                 metrics.inc "subscribe.failed.#{baseChannel}"
-                # clear state
-                clientChannelMap.delete(channel)
+                # clear state on error, following subscribes should retry
+                # new subscribe requests can overtake, skip cleanup then
+                if clientChannelMap.get(channel) is subscribePromise
```
briangough, line 38: Ah yes, that is an important check that I missed.

test/unit/coffee/ChannelManagerTests.coffee
das7pad: OK, so we have to handle rejects on every attached promise. This is unexpected at first, but makes sense in that some arbitrary handler should always handle success and failure. (This may have catched #134.)

<details>

<pre>
> reject=0; p = new Promise((_, r) => {reject=r})
Promise { <pending> }
> p.catch((reason)=>console.log({reason}))
Promise { <pending> }
> p.then()
Promise { <pending> }
> p.then()
Promise { <pending> }
> p.then()
Promise { <pending> }
> reject('boom')
undefined
> { reason: 'boom' }
(node:19102) UnhandledPromiseRejectionWarning: boom
(node:19102) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:19102) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:19102) UnhandledPromiseRejectionWarning: boom
(node:19102) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:19102) UnhandledPromiseRejectionWarning: boom
(node:19102) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 3)
</pre>
</details>

das7pad: Done and force pushed.

das7pad:
Super seeded by #137 and #139

PR #133, opened 2020-05-12
[ChannelManager] enforce a strict sequence of redis (un)subscribes

das7pad:
### Description

Chained onto #132 

This PR implements a strict sequence of subscribe and unsubscribe calls.

It is a safe guard for future work of arbitrary cleanup calls.

Subscribe waits for unsubscribe requests. Unsubscribe waits for
 subscribe requests. Unsubscribe breaks the chain.

Additionally I added shortcuts into the unsubscribe process to skip the redis request
 when we detect a race/overtaking request. This is important for the future delayed
 unsubscribe calls -- they should not block any new subscribes or do duplicate work.

1st subscribe -> 1st unsubscribe -> 2nd subscribe -> 2nd unsubscribe

We can skip the 1st unsubscribe, when the 2nd subscribe came in before the 1st subscribe completed. We can also skip the 1st unsubscribe when the 2nd unsubscribe came in before the 1st subscribe completed.

#### Review

I could not find a simple way to tell whether a Promise completed already, so there is a home grown solution which lets a given Promise race against another that completes 1ms later.
I am happy to replace it with something simpler. Moving to a new coffee-script version or decaff would enable `async function`s, but we are not there yet.

Regarding the acceptance tests:

`test/acceptance/coffee/PubSubRace.coffee -> when the client disconnects before joinDoc completes`:

Due to different code paths in socket.io for `joinProject` and `disconnect` requests from the client, we are very likely not getting into a race (two vs one process.nextTick calls).

#### Related Issues / PRs

#### Potential Impact

High. clients could be connected without getting any updates.

#### Manual Testing Performed

- added lots of unit tests
- a new acceptance test suite
- open lots of editor tabs, they should not see each other

### DIFF
--- a/app/coffee/ChannelManager.coffee
+++ b/app/coffee/ChannelManager.coffee
@@ -40,13 +40,10 @@ module.exports = ChannelManager =
                         clientChannelMap.delete(channel)
                 return p
 
-            if clientChannelMapTearDown.has(channel)
-                # wait for the unsubscribe request to complete
-                unsubscribePromise = clientChannelMapTearDown.get(channel)
-                subscribePromise = unsubscribePromise.finally(actualSubscribe)
-
-            else
-                subscribePromise = actualSubscribe()
+            # wait for the unsubscribe request to complete
+            unsubscribePromise = Promise.resolve(clientChannelMapTearDown.get(channel))
+            # unsubscribePromise never rejects
+            subscribePromise = unsubscribePromise.then(actualSubscribe)
 
             clientChannelMap.set(channel, subscribePromise)
             logger.log {channel}, "subscribed to new channel"
@@ -72,19 +69,21 @@ module.exports = ChannelManager =
                     clientChannelMapTearDown.delete(channel)
                     return Promise.resolve()
 
-                p = rclient.unsubscribe channel
-                p.catch (err) ->
-                    logger.error {channel, err}, "failed to unsubscribed from channel"
-                    metrics.inc "unsubscribe.failed.#{baseChannel}"
-                p.finally () ->
+                # nothing is attaching to this Promise, catch any errors
+                p = rclient.unsubscribe(channel)
+                .finally () ->
                     # new unsubscribe requests can overtake, skip cleanup
                     if clientChannelMapTearDown.get(channel) is unsubscribePromise
                         clientChannelMapTearDown.delete(channel)
+                .catch (err) ->
+                    logger.error {channel, err}, "failed to unsubscribed from channel"
+                    metrics.inc "unsubscribe.failed.#{baseChannel}"
                 return p
 
             # wait for the subscribe request to complete
             subscribePromise = clientChannelMap.get(channel)
-            unsubscribePromise = subscribePromise.finally(actualUnsubscribe)
+            # subscribePromise can reject, unsubscribe in any case
+            unsubscribePromise = subscribePromise.then(actualUnsubscribe, actualUnsubscribe)
             clientChannelMapTearDown.set(channel, unsubscribePromise)
 
             # new subscribe requests must wait for the unsubscribe request to
--- a/app/coffee/ChannelManager.coffee
+++ b/app/coffee/ChannelManager.coffee
@@ -3,6 +3,7 @@ metrics = require "metrics-sharelatex"
 settings = require "settings-sharelatex"
 
 ClientMap = new Map() # for each redis client, store a Map of subscribed channels (channelname -> subscribe promise)
+ClientMapTearDown = new Map() # as ClientMap but tracks unsubscribe requests
 
 # Manage redis pubsub subscriptions for individual projects and docs, ensuring
 # that we never subscribe to a channel multiple times. The socket.io side is
@@ -14,8 +15,12 @@ module.exports = ChannelManager =
         # return an empty map for the client.
         ClientMap.get(rclient) || ClientMap.set(rclient, new Map()).get(rclient)
 
+    getClientMapTearDownEntry: (rclient) ->
+        ClientMapTearDown.get(rclient) || ClientMapTearDown.set(rclient, new Map()).get(rclient)
+
     subscribe: (rclient, baseChannel, id) ->
         clientChannelMap = @getClientMapEntry(rclient)
+        clientChannelMapTearDown = @getClientMapTearDownEntry(rclient)
         channel = "#{baseChannel}:#{id}"
         # we track pending subscribes because we want to be sure that the
         # channel is active before letting the client join the doc or project,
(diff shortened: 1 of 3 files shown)


app/coffee/ChannelManager.coffee
```diff
@@ -25,33 +30,71 @@ module.exports = ChannelManager =
...
+                    # clear state on error, following subscribes should retry
+                    # new subscribe requests can overtake, skip cleanup then
+                    if clientChannelMap.get(channel) is subscribePromise
+                        clientChannelMap.delete(channel)
+                return p
+
+            if clientChannelMapTearDown.has(channel)
+                # wait for the unsubscribe request to complete
+                unsubscribePromise = clientChannelMapTearDown.get(channel)
+                subscribePromise = unsubscribePromise.finally(actualSubscribe)
```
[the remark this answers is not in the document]
das7pad: I am not particularly interested in the return value of the subscribe call, but just the sequence of the executions. It is flawed with rejections however: in your example `x` is rejected when `a` is rejected. I will give this some thinking.

<details>

```js
let downRes, downRej, subRes, subRej

console.log('---resolve down---')

down = new Promise((res, rej) => {downRes=res; downRej=rej})
down.then(()=>console.log('down completed'))
sub = new Promise((res, rej) => {subRes=res; subRej=rej})
sub.then(()=>console.log('sub completed'))
mapEntry = down.finally(sub)
mapEntry.finally(()=>console.log('mapEntry completed'))
console.log('three pending', {down, sub, mapEntry})
downRes()
console.log('two pending', {down, sub, mapEntry})
subRes()
setTimeout(() => console.log('none pending', {down, sub, mapEntry}))

setTimeout(() => {// let the other test finish
console.log('---reject down---')

down = new Promise((res, rej) => {downRes=res; downRej=rej})
down.catch(()=>console.log('down completed'))
sub = new Promise((res, rej) => {subRes=res; subRej=rej})
sub.then(()=>console.log('sub completed'))
mapEntry = down.finally(sub)
mapEntry.finally(()=>console.log('mapEntry completed'))
console.log('three pending', {down, sub, mapEntry})
downRej()
console.log('two pending', {down, sub, mapEntry})
subRes()
setTimeout(() => console.log('none pending', {down, sub, mapEntry}))
}, 10)
``` 

renders as
```
$ nvm run 10.19.0 promise_finally.js 
Running node v10.19.0 (npm v6.13.4)
---resolve down---
three pending { down: Promise { <pending> },
  sub: Promise { <pending> },
  mapEntry: Promise { <pending> } }
two pending { down: Promise { undefined },
  sub: Promise { <pending> },
  mapEntry: Promise { <pending> } }
down completed
sub completed
mapEntry completed
none pending { down: Promise { undefined },
  sub: Promise { undefined },
  mapEntry: Promise { undefined } }
---reject down---
three pending { down: Promise { <pending> },
  sub: Promise { <pending> },
  mapEntry: Promise { <pending> } }
two pending { down: Promise { <rejected> undefined },
  sub: Promise { <pending> },
  mapEntry: Promise { <pending> } }
down completed
sub completed
mapEntry completed
(node:3524) UnhandledPromiseRejectionWarning: undefined
(node:3524) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:3524) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
none pending { down: Promise { <rejected> undefined },
  sub: Promise { undefined },
  mapEntry: Promise { <rejected> undefined } }
``` 
</details>

das7pad:
> From a style point of view I think it might be cleaner if both functions were written symmetrically as

This won't work for subscribe, as we may not have a pending unsubscribe to chain onto.

briangough:
> > From a style point of view I think it might be cleaner if both functions were written symmetrically as
> 
> This won't work for subscribe, as we may not have a pending unsubscribe to chain onto.

If you use `Promise.resolve(...get pending from map here..)` and the pending result is udnefined then it will do the right thing.

test/unit/coffee/ChannelManagerTests.coffee
```diff
@@ -17,104 +17,155 @@ describe 'ChannelManager', ->
...
-					return new Promise (resolve, reject) ->
-						setTimeout((() -> reject(new Error("some redis error"))), 3)
+				@rclient.subscribe = sinon.stub()
+					.onFirstCall().rejects(new Error("some redis error"))
+					.onSecondCall().resolves()
 				@first = @ChannelManager.subscribe @rclient, "applied-ops", "1234567890abcdef"
 				# ignore error
 				@first.catch((()->))
-				@ChannelManager.getClientMapEntry(@rclient).get("applied-ops:1234567890abcdef").should.equal @first
+				expect(@ChannelManager.getClientMapEntry(@rclient).get("applied-ops:1234567890abcdef")).to.equal @first
```
das7pad, line 66: For some reason, this was starting to fail. Adding a console.log showed a pending promise for the map entry, but `should` was `undefined`.

test/unit/coffee/ChannelManagerTests.coffee
```diff
@@ -17,104 +17,155 @@ describe 'ChannelManager', ->
...
+			beforeEach (done) ->
 				@rclient.subscribe = sinon.stub().resolves()
 				@ChannelManager.subscribe @rclient, "applied-ops", "1234567890abcdef"
-				@rclient.subscribe = sinon.stub().resolves()  # discard the original stub
 				@ChannelManager.subscribe @rclient, "applied-ops", "1234567890abcdef"
+				setTimeout done
 
 			it "should not subscribe to the redis channel", ->
-				@rclient.subscribe.called.should.equal false
+				@rclient.subscribe.callCount.should.equal 1
```
das7pad, line 36: Replacing the stub causes a race condition with the _always delayed_ `actualSubscribe` execution. `rclient.subscribe` is a single stub instance now. There is a similar change right below.

das7pad:
I do agree with you that the optimization is complicating the unsubscribe process. The simple approach is coming at the cost of subscribing again for every collaborator on the same pod.

Shall we target this PR onto v0 and test the impact on redis under production load first?

briangough:
> I do agree with you that the optimization is complicating the unsubscribe process. The simple approach is coming at the cost of subscribing again for every collaborator on the same pod.
> 
> Shall we target this PR onto v0 and test the impact on redis under production load first?

Agreed.

das7pad:
I will prepare a second PR with your proposed subscribe/unsubscribe + error handling.

das7pad:
Super seeded by #137 and #139

PR #135, opened 2020-05-12
[WebsocketController] skip leaveProject when joinProject didn't complete

das7pad:
#### Description

We are seeing random spikes in the metrics from clients connecting, but never actually authenticating/joining. Upon disconnect we are calling leaveProject to cleanup, which may not be actually needed, when the client did not call joinProject yet.

I attached a `joinedProject` attribute to the client when the web api call succeeded. Only there after we are actually populating client params and joining rooms.

#### Review

Putting the `joinedProject` flag into the regular client params would require indent changes to get the value back in an async way.

Alternative approach: We could move `metrics.inc "editor.leave-project"` into `Utils.getClientAttributes` and use a regular param then.

#### Potential Impact

High, missing cleanup could lead to memory leaks.

#### Manual Testing Performed

- added unit test

### DIFF
--- a/app/coffee/WebsocketController.coffee
+++ b/app/coffee/WebsocketController.coffee
@@ -51,22 +51,13 @@ module.exports = WebsocketController =
 	# is determined by FLUSH_IF_EMPTY_DELAY.
 	FLUSH_IF_EMPTY_DELAY: 500 #ms
 	leaveProject: (io, client, callback = (error) ->) ->
-		metrics.inc "editor.leave-project"
 		Utils.getClientAttributes client, ["project_id", "user_id"], (error, {project_id, user_id}) ->
 			return callback(error) if error?
+			return callback() unless project_id # client did not join project
 
+			metrics.inc "editor.leave-project"
 			logger.log {project_id, user_id, client_id: client.id}, "client leaving project"
 			WebsocketLoadBalancer.emitToRoom project_id, "clientTracking.clientDisconnected", client.id
-
-			# bail out if the client had not managed to authenticate or join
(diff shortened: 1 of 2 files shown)


app/coffee/WebsocketController.coffee
```diff
@@ -51,6 +54,7 @@ module.exports = WebsocketController =
 	# is determined by FLUSH_IF_EMPTY_DELAY.
 	FLUSH_IF_EMPTY_DELAY: 500 #ms
 	leaveProject: (io, client, callback = (error) ->) ->
+		return callback(new Error('not joined')) unless client.joinedProject
```
briangough, line 57: let's move the metric into the getClientAttributes and add a check on `project_id` instead of adding a new `joinedProject` attribute.  

I think we can skip returning an error and silently return from leaveProject() if the client never joined, that makes the rest of the code simpler.  

If you put the metric after this check then we can see the difference between `editor.leave-project` and `socket.io-disconnect` to tell how many clients are disconnecting without joining.  Alternatively we can have a separate metric for it if you think it's worth tracking.
```
return callback() if !project_id?  # client did not join project
metrics.inc "editor.leave-project"
```

das7pad: This now breaks other unit tests. I also found flawed code that handed the missing user_id and project_id at a later stage. See commit comment:

 - user_id bailout

   There is a check on a completed joinProject call now. It will always
    set a user_id, see Router.coffee which has a fallback `{_id:"..."}`.

 - late project_id bailout

   WebsocketLoadBalancer.emitToRoom will not work without a project_id.
   We have to bail out before the call.

PR #137, opened 2020-05-18
[misc] backport strict sequence of redis (un)subscribe

das7pad:
### Description

Backport of the simplified strict-sequence from #133

Please see the discussion over there for more details.

Chained onto #136 for promise support with stubs.

#### Related Issues / PRs

#133

#### Potential Impact

High, every `joinProject` and `joinDoc` triggers a subscribe redis request -- previously only the first collaborator on the same pod triggered the subscribe, following users joining the same project/doc on a given pod used the existing subscription.

#### Manual Testing Performed

- added new acceptance tests with a race of joinDoc vs leaveDoc requests
- extended existing acceptance tests to check redis subscribes
- added unit tests

#### Metrics and Monitoring

Watch redis dashboard and upstream redis dashboard for CPU usage.

PR #140, opened 2020-05-20
[misc] stop processing requests as we detect a client disconnect

das7pad:
### Description

Chained onto https://github.com/overleaf/real-time/pull/139 which merges v0 (including a pub/sub race test suite)

This PR fixes racing join requests and disconnect events. The processing of disconnect events skips a `process.nextTick` call (2 vs 3 calls) and hence we will process disconnect before a join request (or other rpc call).

- joinProject
   -  bails out at start, after web api call
   - previous impact: could leave dangling pub/sub subscribes

- joinDoc

  - bails out at start, after redis pub/sub join (1), after doc-updater call
  - previous impact: could leave dangling pub/sub subscribes, e.g. see [this failed build](https://console.cloud.google.com/cloud-build/builds/604379c6-d5d0-4f49-b3b1-ed0f97989b72;step=2?project=overleaf-ops)

- updateClientPosition
   - bails out at start
   - previous impact: could create ghost clients

- getConnectedUsers
   - bails out at start
   - previous impact: multiple useless redis calls

- applyOtUpdate

   - I chose to process the update anyways, even tho we can not signal a processing error back.
   - bail out when notifying the client about a too large update with a 100ms delay
   - previous impact: useless processing on dead client

(1) this does not need any extra handling: we join the socket.io room before calling into redis, hence the disconnect will pick up the existing socket.io room membership and leave it (it will wait for subscribe to complete first #137).

#### Screenshots

Race condition flagged in CI https://console.cloud.google.com/cloud-build/builds/604379c6-d5d0-4f49-b3b1-ed0f97989b72;step=2?project=overleaf-ops

#### Related Issues / PRs

#### Potential Impact

High, join requests could fail for unstable connections. We [create a new socket.io connection/client per (re)connect ](https://github.com/overleaf/web/blob/4c252fb81f495204136d32e5fe3730884518d746/frontend/js/ide/connection/SocketIoShim.js#L140-L163) and the [reconnect feature of socket.io is turned off](https://github.com/overleaf/web/blob/4c252fb81f495204136d32e5fe3730884518d746/frontend/js/ide/connection/ConnectionManager.js#L131), so this fear is not warranted.

#### Manual Testing Performed

- full coverage via unit tests
- new acceptance test suite which covers different races (joinProject, joinDoc, updateClientPosition)
  Given that we can not force the joinDoc and updateClientPosition race condition, I retry the tests multiple times, which increases the test suite run time by about 10s.
- there is an existing PubSubRace test suite 

#### Metrics and Monitoring

`disconnected_join_project`
`disconnected_join_doc`
`disconnected_otUpdateError`

### DIFF
--- a/app/coffee/WebsocketController.coffee
+++ b/app/coffee/WebsocketController.coffee
@@ -15,11 +15,18 @@ module.exports = WebsocketController =
 	PROTOCOL_VERSION: 2
 
 	joinProject: (client, user, project_id, callback = (error, project, privilegeLevel, protocolVersion) ->) ->
+		if not client.connected
+			metrics.inc('disconnected_join_project')
+			return callback()
+
 		user_id = user?._id
 		logger.log {user_id, project_id, client_id: client.id}, "user joining project"
 		metrics.inc "editor.join-project"
 		WebApiManager.joinProject project_id, user, (error, project, privilegeLevel, isRestrictedUser) ->
 			return callback(error) if error?
+			if not client.connected
+				metrics.inc('disconnected_join_project')
+				return callback()
 
 			if !privilegeLevel or privilegeLevel == ""
 				err = new Error("not authorized")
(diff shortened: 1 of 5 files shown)


app/coffee/WebsocketController.coffee
```diff
@@ -15,11 +15,18 @@ module.exports = WebsocketController =
 	PROTOCOL_VERSION: 2
 
 	joinProject: (client, user, project_id, callback = (error, project, privilegeLevel, protocolVersion) ->) ->
+		if not client.connected
+			metrics.inc('disconnected_join_project')
+			return callback()
```
briangough, line 20: To match the other metrics how about `metrics.inc('editor.join-project.disconnected', 1, {status: 'at-start'})` and below `metrics.inc('editor.join-project.disconnected', 1, {status: 'after-web-api-call'})`?  Similarly for the `joinDoc` method.

PR #158, opened 2020-06-19
[WIP] plain ws

das7pad:
### Description

For https://github.com/overleaf/issues/issues/3195

Currently based onto v2 to get the `ol_context` logic in.

The diff is more verbose that it has to be -- mostly due to the removal of the `io` parameter across the service.

### DIFF
--- a/app/coffee/RoomManager.coffee
+++ b/app/coffee/RoomManager.coffee
@@ -5,6 +5,8 @@ metrics = require "metrics-sharelatex"
 IdMap = new Map() # keep track of whether ids are from projects or docs
 RoomEvents = new EventEmitter() # emits {project,doc}-active and {project,doc}-empty events
 
+RoomClients = new Map() # 'room' -> [clients...]
+
 # Manage socket.io rooms for individual projects and docs
 #
 # The first time someone joins a project or doc we emit a 'project-active' or
@@ -47,10 +49,10 @@ module.exports = RoomManager =
         return RoomEvents
 
     joinEntity: (client, entity, id, callback) ->
-      @_clientsInRoom client, id, (beforeCount) ->
+        beforeCount = @countClientsInRoom(id)
         # client joins room immediately but joinDoc request does not complete
         # until room is subscribed
-        client.join id
+        @_clientJoinRoom client, id
         # is this a new room? if so, subscribe
         if beforeCount == 0
             logger.log {entity, id}, "room is now active"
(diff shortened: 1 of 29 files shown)


app/coffee/RoomManager.coffee
```diff
@@ -88,41 +90,36 @@ module.exports = RoomManager =
...
-        # The adapter will get updated immediately from joining, but the
-        #  client local state (.rooms) is updated via process.nextTick.
-        # Given that NodeJS can process multiple network events in the same
-        #  event loop cycle, we can join a project and disconnect in the same
-        #  cycle. We would not pick up the room (we just joined) for leaving.
-        roomRegistry = client.server.sockets.adapter.sids
-        return [] unless roomRegistry.hasOwnProperty(client.id)
-        rooms = (room for room in Object.keys(roomRegistry[client.id]) when room isnt client.id)  # exclude the client socket entry
-        return rooms
+    getClientsInRoomPseudoAsync: (room, cb) ->
```
gh2k, line 124: `getClientsInRoomSnapshot` might be a better name for this?

das7pad: Note that this name is not new, only the signature changed (like many others -- I should have kept the `io` parameter for a simpler review). I agree that the name is not perfect. A refactoring in the calling logic to a synchronous call would eliminate it.

app/coffee/WebsocketServer.js
```diff
@@ -0,0 +1,435 @@
...
+  app.get(`${Settings.socketIoPath}/socket.io.js`, function(req, res) {
+    res.contentType('application/javascript')
+    res.send('window.io="plain"')
+  })
+
+  const POLLING_PREFIX = `${Settings.socketIoPath}/deprecated-polling`
+  app.use(`${POLLING_PREFIX}/`, addCORSHeaders)
+  app.post(`${POLLING_PREFIX}/new`, function createNewSession(req, res) {
+    const polling = new PollingWebSocketShim()
+    const client = new Client(polling, clientMap)
```
jdleesmiller, line 89: This constructor has two arguments, but the `Client` class in this file only appears to take one. Is that right?

das7pad: Yes, the second parameter is an artifact from a previous iteration and can get removed.

_The `Client` wrapper was tasked to set up the client tracking (insert into the `clientMap` and register a `disconnect` handler to remove it). This logic resides [in the session start](https://github.com/overleaf/real-time/blob/b83405de377b448f9d1a64d17a78552109ee8e96/app/coffee/WebsocketServer.js#L59-L60) now._

PR #162, opened 2020-06-30
[misc] fix unit tests

das7pad:
### Description

During the decaff cleanup https://github.com/overleaf/real-time/pull/163 I stumbled across a few _broken_ tests and a minor inconsistency in regards to sync/async behavior (calling a callback synchronous in one branch and after a redis call in another branch).

This PR addresses these for separate review of the decaff cleanup changes.

This PR is chained onto the decaff-PR #161 

#### Related Issues / PRs

https://github.com/overleaf/real-time/pull/161
https://github.com/overleaf/real-time/pull/163

#### Potential Impact

Low. The only app change is in the code path of token-access users. The acceptance include cases for these.

PR #163, opened 2020-06-30
[misc] decaff cleanup

das7pad:
### Description

This PR cleans up the code base after the decaffeinate process #161 . I've left the camelCase rule disabled as it produces a too verbose diff. Instead I will create a follow-up PR which replaces all snake_case usages across all files -- including the `client.ol_context`. 

Chained onto #162 for fixed in the unit tests, which in turn is chained onto the decaff-PR #161.

#### Related Issues / PRs

https://github.com/overleaf/real-time/pull/161
https://github.com/overleaf/real-time/pull/162

### Review

I started with a commit per file and moved to a commit per rule/topic for the last 6 file (which was very slow, but here we are).

#### Potential Impact

High, all of real-time. 

#### Manual Testing Performed

- run unit/acceptance tests
- open editor in dev-env

### DIFF
--- a/app.js
+++ b/app.js
@@ -86,7 +86,7 @@ const rclient = require('redis-sharelatex').createClient(
   Settings.redis.realtime
 )
 
-const healthCheck = (req, res, next) =>
+function healthCheck(req, res, next) {
   rclient.healthCheck(function (error) {
     if (error) {
       logger.err({ err: error }, 'failed redis health check')
@@ -99,7 +99,7 @@ const healthCheck = (req, res, next) =>
       res.sendStatus(200)
     }
   })
-
+}
 app.get('/health_check', healthCheck)
 
 app.get('/health_check/redis', healthCheck)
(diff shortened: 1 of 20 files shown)


jdleesmiller:
I'm not suggesting we change it here, but it may be worth noting for future reference that the (internal) [dev manual](https://github.com/overleaf/developer-manual/blob/9e3fe5ed6eb5b636829c5baba01f95f9092218de/code/creating-a-decaf-cleanup-PR.md) recommends committing the decaf cleanup rule-by-rule instead of file-by-file, because the former may be (?) easier to review.

das7pad:
> I'm not suggesting we change it here, but it may be worth noting for future reference that the (internal) [dev manual](https://github.com/overleaf/developer-manual/blob/9e3fe5ed6eb5b636829c5baba01f95f9092218de/code/creating-a-decaf-cleanup-PR.md) recommends committing the decaf cleanup rule-by-rule instead of file-by-file, because the former may be (?) easier to review.

This change would result in 3 commits then (return, IIFE, Array.from) -- I'm not sure this really helps in the review process.

https://github.com/overleaf/real-time/pull/163/commits/6d3526c37f5a961a62139a6663d6f6d836f5782a#diff-3d8338fcbae0fc03f424012d8ac32037R42-R45

cc @briangough I will probably request a review from you. Which approach do you prefer?

briangough:
> cc @briangough I will probably request a review from you. Which approach do you prefer?

No need to change the existing commits for this PR, any future ones can be per-rule.

das7pad:
> No need to change the existing commits for this PR, any future ones can be per-rule.

Snap, I've continued in the meantime -- I'll commit per rule for the last 4 modules/files. I hope that's OK?

PR #66, opened 2019-07-25
Patched EventEmitter for socket.io compatibility with Node >= 7

mserranom:
### Description

As part of Server Pro revival we need to be able to run all the services under the same node version. This is the only service that fails to bootstrap, and this patch fixes it. I haven't tested the behaviour yet, only the bootstrap.

#### Related Issues / PRs

Contributes to https://github.com/overleaf/issues/issues/2043

#### Manual Testing Performed

- [x] service bootstraps locally with Node 10
- [x] service continues working under `dev-environment`

#### Deployment Checklist

- [ ] Check service continues working as usual in staging
- [ ] Check service continues working as usual in production

PR #168, opened 2020-07-08
[misc] use overleaf/socket.io and overleaf/socket.io-client forks

das7pad:
### Description

For https://github.com/overleaf/issues/issues/3245

#### Related Issues / PRs

See below.

#### Potential Impact

High. Includes changes to the core parser logic server side and rpc response parser client side.

##### client parser changes:

- [x] https://github.com/overleaf/issues/issues/3199

- break a project

```js
fetch(`/project/${window.project_id}/doc/__INSERT_A_VALID_DOC_ID_FOR_THE_PROJECT__/rename`, {
  method: 'POST',
  headers: {'Content-Type': 'application/json'},
  body: JSON.stringify({
    name: 'pre_' + String.fromCharCode(8233) + '_after.tex',
    _csrf: window.csrfToken
  })
})
```

- reload the editor
- it should complete the init (previously it failed parsing the `joinProject` payload)

##### server deep dependency import changes

- [x] https://github.com/overleaf/real-time/pull/66

Real-time would not startup when this patch is not applied and using node version 7+.

##### server parser changes - websocket

- [ ] https://github.com/overleaf/issues/issues/2017

cc @briangough Is there an easy way to test this?

##### server parser changes - misc

- [x] See the linked ticket https://github.com/overleaf/issues/issues/3050


briangough:
> ##### server parser changes - websocket
> * [ ]  [overleaf/issues#2017](https://github.com/overleaf/issues/issues/2017)
> 
> cc @briangough Is there an easy way to test this?

Perhaps adding an acceptance test which sends packets with increasing amounts of padding?  But to be honest maybe it's not worth it - I think we can just assume that it works as we have been using it in production.

das7pad:
> But to be honest maybe it's not worth it - I think we can just assume that it works as we have been using it in production.

I'll go with that for now.

PR #170, opened 2020-07-20
[minimal] prepare a minimal version of real-time for ws. testing

das7pad:
### Description

For https://github.com/overleaf/issues/issues/3244

It's a copy of app.js without:
 - the actual Router
 - real-time capability
 - continuous pub/sub traffic
 - proper draining on shutdown

It has a very basic rpc on the connection tracking endpoint which just
 echos back an empty list of clients. The regular events for signaling
 invalid sessions, accepted connections, reties are copied from the core
 real-time service. The test kicks off clients after 30s. There are lots
 of metrics for various errors and the general tracking of currently
 connected clients.

Logging and metrics are namespaced into `real-time-minimal.`.

We should be able to use the same docker image and specify a different
 CMD in k8s to run node on the minimal.js file instead.

I chose a random unassigned port 9311 for the service.

#### Related Issues / PRs

https://github.com/overleaf/issues/issues/3244

#### Manual Testing Performed

See https://github.com/overleaf/issues/issues/3244#issuecomment-661157700


gh2k:
Parking, see: https://github.com/overleaf/google-ops/pull/1158#issuecomment-724653846

PR #181, opened 2020-08-13
[misc] forcefully disconnect from shutdown process

das7pad:
### Description

For https://github.com/overleaf/issues/issues/3419

The shutdown process asks clients to reconnect, but some clients may be broken/ignore the message.
This leaves the real-time processes in an unhealthy state indefinitely.

This PR waits for all clients to get asked for a reconnect and after a configurable delay `GRACEFUL_RECONNECT_TIMEOUT_MS` it disconnects them forcefully.

The default delay is 45s + 30s. The 45s are [coming from the frontend](https://github.com/overleaf/web/blob/eb84e0d4eb2ea83e9f31753b10297e71ae3d02c2/frontend/js/ide/connection/ConnectionManager.js#L32) -- it permits users to keep editing. I added 30s to allow users with a bad internet connectivity to receive the graceful reconnect message and flush their send queue.

#### Related Issues / PRs

https://github.com/overleaf/issues/issues/3419

#### Potential Impact

Low. Affects the shutdown process only.

#### Manual Testing Performed

- add signal forwarding to the node-process, skip nodemon for real-time:

<details>

```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index f12e4fe91..f6dc76611 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -123,6 +123,7 @@ services:
     extends:
       file: common.yml
       service: common
+    command: node --expose-gc app.js
     volumes:
       - ./real-time:/app:cached
     ports:
```
</details>

- `dev-env$ bin/up core`
- open a project in the editor
- in the browser console, create a few new headless clients `io.connect(null, {'force new connection': 1})`
- tail logs `dev-env$ docker-compose logs --tail=10 -f real-time`
- curl the app health check: `$ curl 127.0.0.1:3026/health_check` -> `OK`
- shutdown real-time with a long timeout `dev-env$ docker-compose stop -t900 real-time`
- see the reconnectGracefully requests in the log
- see the number of headless clients in the log message of periodic shutdownCleanly polling

   > `"connectedClients":30,"msg":"clients still connected, not shutting down yet"`
- eventually (after about 45s+30s) see `forcefully disconnecting stale clients` with the client ids of the headless clients

  > `"staleClients":["stlPGKhEqVYBDGMouRnE", ... ,"Xgb5CrjJxMJgB9N_uRnh"],"msg":"forcefully disconnecting stale clients"`
- curl the app health check: `$ curl 127.0.0.1:3026/health_check` -> `Service Unavailable`
- curl the redis health check: `$ curl 127.0.0.1:3026/health_check/redis` -> `OK`
- eventually (<30s) `shutdownCleanly` emits `no clients connected, exiting` and the process exits

### DIFF
--- a/app.js
+++ b/app.js
@@ -155,7 +155,20 @@ function drainAndShutdown(signal) {
         { signal },
         `received interrupt, starting drain over ${shutdownDrainTimeWindow} mins`
       )
-      DrainManager.startDrainTimeWindow(io, shutdownDrainTimeWindow)
+      DrainManager.startDrainTimeWindow(io, shutdownDrainTimeWindow, () => {
+        setTimeout(() => {
+          const staleClients = io.sockets.clients()
+          if (staleClients.length !== 0) {
+            logger.warn(
+              { staleClients: staleClients.map((client) => client.id) },
+              'forcefully disconnecting stale clients'
+            )
+            staleClients.forEach((client) => {
+              client.disconnect()
+            })
+          }
+        }, Settings.gracefulReconnectTimeoutMs)
+      })
       shutdownCleanly(signal)
     }, statusCheckInterval)
   }
--- a/app.js
+++ b/app.js
@@ -98,7 +98,16 @@ function healthCheck(req, res) {
     }
   })
 }
-app.get('/health_check', healthCheck)
+app.get(
+  '/health_check',
+  (req, res, next) => {
+    if (Settings.shutDownComplete) {
+      return res.sendStatus(503)
+    }
+    next()
+  },
+  healthCheck
+)
 
 app.get('/health_check/redis', healthCheck)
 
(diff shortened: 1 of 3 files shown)


app.js
```diff
@@ -155,7 +164,22 @@ function drainAndShutdown(signal) {
...
+      DrainManager.startDrainTimeWindow(io, shutdownDrainTimeWindow, () => {
+        setTimeout(() => {
+          const staleClients = io.sockets.clients()
+          if (staleClients.length !== 0) {
+            logger.warn(
+              { staleClients: staleClients.map((client) => client.id) },
+              'forcefully disconnecting stale clients'
+            )
+            staleClients.forEach((client) => {
+              client.disconnect()
```
henryoswald, line 176: Would this code make more sense in the drain manager?

das7pad: I would like to keep the DrainManager as is -- the DrainManager is more-or-less stateless and the drain process can get stopped/adjusted at any time, whereas there is no way back with the newly proposed marking as unhealthy at the end. Implementing a ShutdownManager with all the shutdown logic may be a something for a future task?

PR #198, opened 2020-11-11
[misc] bump @overleaf/redis-wrapper to version 2.0.0

das7pad:
### Description

For https://github.com/overleaf/issues/issues/3650

- removed support for redis-sentinel
- redis-wrapper was rewritten in plain ES and promisified
  - mostly internal changes
  - health check supports `await` natively
  - `multi().exec()` supports `await` and unpacks the `ioredis` result
- package renamed as `@overleaf/redis-wrapper`
   - requires changed imports
- added `peerDependency` of `@overleaf/o-error@3.1.0`
   - requires an added/updated production dependency on o-error
- bring back support for redis-cluster

#### Related Issues / PRs

For https://github.com/overleaf/issues/issues/3650

https://github.com/overleaf/redis-wrapper/pull/16

#### Potential Impact

High. The health check has been rewritten, so was the ioredis result unpacking.

#### Manual testing performed

- Health check: simulate redis failures (unavailable, corrupted read)
- I tested the services locally in dev-environment
   - dashboard
   - editor
   - edits
- I deployed from the PR branch into staging (all but analytics)
   - same as for local testing
   - dropbox sync tested in both directions
- The rewrite was upstreamed from my fork, and most of it was running since
   the end of August (ES/o-error since April) in my infra with a redis-cluster.


app.js
```diff
@@ -94,7 +94,7 @@ app.get('/debug/events', function (req, res) {
   res.send(`debug mode will log next ${Settings.debugEvents} events`)
 })
 
-const rclient = require('redis-sharelatex').createClient(
+const rclient = require('@overleaf/redis-wrapper').createClient(
```
gh2k, line 97: Looks like we've already required this once. Would this be cleaner?
```suggestion
const rclient = redis.createClient(
```

gh2k (approved):
Nice :-)

Good to see this landing. Just one suggestion that doesn't require re-review.

app.js
das7pad: Prettier will likely complain about this. Something for a follow-up.

PR #205, opened 2021-02-08
Pending update shard

henryoswald:
### Description
The same as the document update pr, this shards the 'pending-updates-list' key.

I've set the key in settings to 1 for the time being, I can modify this again in a later pr.

 

#### Related Issues / PRs
https://github.com/overleaf/document-updater/pull/156
https://github.com/overleaf/issues/issues/3955

#### Potential Impact
High - is ops are not processed it will cause the editor to break

#### Manual Testing Performed

- [ ]
- [ ]

### Deployment
I've not worked out a good deployment strategy yet. I am currently thinking about just increasing the document updater count during a quiet time so there are enough processing the 'pending-updates-list' queue, then quickly flip real time to the new sharded key and finally scale down the document updater.

#### Deployment Checklist

- [ ] Update documentation not included in the PR (if any)
- [ ]

### DIFF
--- a/app/js/DocumentUpdaterManager.js
+++ b/app/js/DocumentUpdaterManager.js
@@ -88,6 +88,15 @@ const DocumentUpdaterManager = {
     })
   },
 
+  _getPendingUpdateListKey() {
+    const shard = _.random(0, settings.pendingUpdateListShardCount)
+    if (shard === 0) {
+      return 'pending-updates-list'
+    } else {
+      return `pending-updates-list-${shard}`
+    }
+  },
+
   queueChange(project_id, doc_id, change, callback) {
     const allowedKeys = [
       'doc',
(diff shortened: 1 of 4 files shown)


app/js/DocumentUpdaterManager.js
```diff
@@ -123,12 +132,18 @@ const DocumentUpdaterManager = {
...
-        if (error) {
-          error = new OError('error pushing doc_id into redis').withCause(error)
+      rclient.rpush(
+        DocumentUpdaterManager._getPendingUpdateListKey(),
+        doc_key,
+        function (error) {
+          if (error) {
+            error = new OError('error pushing doc_id into redis').withCause(
+              error
+            )
```
das7pad, line 142: WDYT about adding the picked queue into the context of the error for debugging?

app/js/DocumentUpdaterManager.js
```diff
@@ -88,6 +88,15 @@ const DocumentUpdaterManager = {
     })
   },
 
+  _getPendingUpdateListKey() {
+    const shard = _.random(0, 10)
```
briangough, line 92: I think there is an off by one error here, as the range of `_.random` is 0 to 10 inclusive: https://lodash.com/docs#random

In docupdater dispatchers are created with `_.times(10, ...)`  which has the range 0-9.

To match docupdater, real-time needs to use `_.random(0,N-1)` if the number of shards is N.

```
> _.times(10)
[
  0, 1, 2, 3, 4,
  5, 6, 7, 8, 9
]
> _.max(_.times(1000, function () { return _.random(0,10) }))
10
```

briangough (approved):

PR #210, opened 2021-03-22
add blue-green deployment support with load-balancer health-checks

briangough:
### Description

This PR allows real-time to use blue-green deployments behind a load-balancer.

To avoid problems with changes to kubernetes services (which can cause downtime) and changes to pod labels (which can revert if a node or pod is recreated) this uses a configmap to allow the real-time pods to signal to the load balancer whether they are available.

The configmap  specifies the current `deployment_colour`, and the pods monitor it via a file.  The  pod colour is defined via an environment variable at start up.  When the pod colour matches the `deployment_colour` then the pod signals to the load balancer that it is available through the load balancer health checks on `/`, otherwise it removes itself from the load balancer by returning `503 Unavailable` on `/`.

#### Screenshots

NA

#### Related Issues / PRs

https://github.com/overleaf/google-ops/pull/1516
https://github.com/overleaf/google-ops/pull/1534

### Review

The open/close mechanism for the blue green flipping is independent of the pod shutdown.  We are essentially recreating the service flip at the pod level.  Configmaps can take up to a minute to update, so to ensure the changeover is graceful I've added a 1 minute delay before old pods are closed.  The opening is immediate.

#### Potential Impact

Real-time connections

#### Manual Testing Performed

- [X]  Test in dev env (Add settings for deployment file and colour to docker config, modify an existing file in the container to change the colour, I used /etc/issue, monitor the load balancer health check on / and see that closing takes 1 minute and opening is immediate).
 
#### Accessibility

NA

### Deployment

Deploy the changes to create the configmap first: 
- [ ] https://github.com/overleaf/google-ops/pull/1516
- [ ] https://github.com/overleaf/google-ops/pull/1534

#### Deployment Checklist

- 

#### Metrics and Monitoring

Realtime connections

### DIFF
--- a/app/js/DeploymentManager.js
+++ b/app/js/DeploymentManager.js
@@ -12,7 +12,7 @@ const deploymentColour = settings.deploymentColour
 var serviceCloseTime
 
 function updateDeploymentStatus(fileContent) {
-  const closed = fileContent && fileContent.indexOf(deploymentColour) === -1
+  const closed = fileContent && !fileContent.includes(deploymentColour)
   if (closed && !settings.serviceIsClosed) {
     settings.serviceIsClosed = true
     serviceCloseTime = Date.now() + 60 * 1000 // delay closing by 1 minute
--- a/app/js/DeploymentManager.js
+++ b/app/js/DeploymentManager.js
@@ -1,0 +1,53 @@
+const logger = require('logger-sharelatex')
+const settings = require('settings-sharelatex')
+const fs = require('fs')
+
+// Monitor a status file (e.g. /etc/real_time_status) periodically and close the
+// service if the file contents don't contain the matching deployment colour.
+
+const FILE_CHECK_INTERVAL = 5000
+const statusFile = settings.deploymentFile
+const deploymentColour = settings.deploymentColour
+
+function updateDeploymentStatus(fileContent) {
+  const closed = fileContent && fileContent.indexOf(deploymentColour) === -1
+  if (closed && !settings.serviceIsClosed) {
+    settings.serviceIsClosed = true
+    logger.warn({ fileContent }, 'closing service')
+  } else if (!closed && settings.serviceIsClosed) {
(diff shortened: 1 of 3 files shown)


app/js/DeploymentManager.js
```diff
@@ -0,0 +1,59 @@
...
+// service if the file contents don't contain the matching deployment colour.
+
+const FILE_CHECK_INTERVAL = 5000
+const statusFile = settings.deploymentFile
+const deploymentColour = settings.deploymentColour
+
+var serviceCloseTime
+
+function updateDeploymentStatus(fileContent) {
+  const closed = fileContent && fileContent.indexOf(deploymentColour) === -1
```
henryoswald, line 15: I think we can do `fileContent.includes(deploymentColour)` now

henryoswald (approved):

app/js/DeploymentManager.js
briangough: Ah, useful.  I will change it.

app/js/DeploymentManager.js
```diff
@@ -0,0 +1,59 @@
...
+const statusFile = settings.deploymentFile
+const deploymentColour = settings.deploymentColour
+
+var serviceCloseTime
+
+function updateDeploymentStatus(fileContent) {
+  const closed = fileContent && !fileContent.includes(deploymentColour)
+  if (closed && !settings.serviceIsClosed) {
+    settings.serviceIsClosed = true
+    serviceCloseTime = Date.now() + 60 * 1000 // delay closing by 1 minute
```
das7pad, line 18: I wonder whether it would make sense to extend the `fileContent` with a value for `serviceCloseTime`. It could allow us to flip all pods at say 12:00 flat so that we get a consistent fail-over between colors.

briangough: I considered using that approach but decided to keep it simple as simple as possible.

PR #211, opened 2021-03-30
Lower log level of 'no project_id found on client' messages

thomas-:
### Description

Reduce log noise by lowering log level of `no project_id found on client` messages.

#### Related Issues / PRs

Papercut 32

### DIFF
--- a/app/js/Router.js
+++ b/app/js/Router.js
@@ -44,12 +44,15 @@ module.exports = Router = {
       metrics.inc('unexpected-arguments', 1, { status: method })
       const serializedError = { message: error.message }
       callback(serializedError)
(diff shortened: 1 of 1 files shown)


gh2k:
I'm tied up for a bit so reassigning to @das7pad

das7pad (approved):
For reference: turning on verbose logging will show `DEBUG` messages as the level is above `TRACE` messages.

https://github.com/overleaf/logger-module/blob/4355a257b0d65180d0e9e647430d15d0627ea9d8/logging-manager.js#L50
https://github.com/trentm/node-bunyan/blob/0ff1ae29cc9e028c6c11cd6b60e3b90217b66a10/lib/bunyan.js#L254-L255

